PR #46, opened 2021-03-09
made function which print message and call it in resolve promise an…

bogdandobak:
…d reject promise


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function printMessage(type, text, check) {
```
solaryasha, line 5:
```suggestion
function printMessage(type, text, shouldDelete) {
```

solaryasha (approved):

PR #169, opened 2021-12-27
Task: First promises

yuran007:
[DEMO LINK](https://yuran007.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,47 @@
+/* eslint-disable promise/param-names */
+/* eslint-disable no-console */
 'use strict';
 
-// write your code here
+const div = document.createElement('div');
+
+document.body.append(div);
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,47 @@
+/* eslint-disable promise/param-names */
+/* eslint-disable no-console */
 'use strict';
 
-// write your code here
+const div = document.createElement('div');
+
+document.body.append(div);
+
+const divFormat = (element, className, message) => {
```
aliceblunt8, line 9: function is an action, so it's better to use verbs
and to make function universal we can add element instead of hard-coding div in its name

```suggestion
const formatElement = (element, className, message) => {
```

aliceblunt8 (approved):
GJ

PR #201, opened 2022-04-26
Solution JS Promise Basic DOM

dmytrosmnnk:
https://dmytrosmnnk.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -29,11 +29,9 @@ const errorHandler = () => {
 };
 
 promise1
-  .then(successHandler)
-  .catch(errorHandler);
+  .then(successHandler);
 
 promise2
-  .then(successHandler)
   .catch(setTimeout(() => {
     errorHandler();
   }, 3000));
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  const div = document.createElement('div');
+
+  body.append(div);
+  div.classList.add('message', 'error-message');
+  div.innerText = 'Promise was rejected!';
+};
+
+promise1
+  .then(successHandler)
+  .catch(errorHandler);
```
marina-tilniak, line 33: this promise will definitely be successful, you don't need catch here

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+
+promise2
+  .then(successHandler)
```
marina-tilniak, line 36: why do you need "then" here?

marina-tilniak (changes requested):

PR #219, opened 2022-06-27
solution

FANTAZER-NURE:


FANTAZER-NURE:
https://FANTAZER-NURE.github.io/js_promise_basic_DOM/

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
 // write your code here
+
+document.querySelector('.logo')
+  .addEventListener('click', () => {
```
aliceblunt8, line 6: you need to create 2 promises and inside each of them add logic for each case, then you need add success and error handlers for both promises (with .then and .catch)

> promise1 should be resolved on a click on .logo.
promise2 promise should be rejected in 3 second

```
   ...
});

   ...
});
```

ArturErmolenko (approved):

PR #225, opened 2022-07-10
Solution promise basic DOM

DimaShm:


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+promise2.catch(error => printMessage(error));
+
+function printMessage(message) {
+  const messageForPrint = document.createElement('div');
+
+  messageForPrint.innerText = message;
+
+  message === 'Promise was resolved!'
+    ? messageForPrint.classList.add('message')
+    : messageForPrint.classList.add('error-message');
```
Bohdan-Maksymiuk, line 27: better not rely on message text
you could pass additional parameter to `printMessage` (boolean error for example with default to false)

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #249, opened 2022-08-10
solution

matvii1:
https://matvii1.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,38 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const divError = `
+  <div class="error-message">
+    Promise was rejected!
+  </div>
+`;
+const divSuccess = `
+  <div class="message">
+    Promise was resolved!
+  </div>
+  `;
+
+const promise1 = new Promise((resolve) => {
+  logo.addEventListener('click', () => {
+    console.log(Promise.resolve());
+    return resolve();
+  })
+})
+  .then(result => {
+    document.body.insertAdjacentHTML("beforeend", divSuccess)
+})
+  .catch(() => {
+    document.body.insertAdjacentHTML("beforeend", divError)
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    return reject();
+  }, 3000)
+})
+.then(() => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+const divError = `
+  <div class="error-message">
+    Promise was rejected!
+  </div>
+`;
+const divSuccess = `
+  <div class="message">
+    Promise was resolved!
+  </div>
+  `;
```
stanislav-topikha, line 13: Those layout are similar, maybe you can create a function that will return layout with args(like `className` and `text`)?

src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+
+const promise1 = new Promise((resolve) => {
+  logo.addEventListener('click', () => {
+    console.log(Promise.resolve());
```
stanislav-topikha, line 17: this shouldn't be here

src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+  setTimeout(() => {
+    return reject();
+  }, 3000)
+})
+.then(() => {
+  document.body.insertAdjacentHTML("beforeend", divSuccess)
+})
+.catch(() => {
+  document.body.insertAdjacentHTML("beforeend", divError)
+});
```
stanislav-topikha, line 38: use you vars, in other case why do you even need them?)

```
promise1
  .then(...)
  .catch(...);

promise2
  .then(...)
  .catch(...);
```

src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+    document.body.insertAdjacentHTML("beforeend", divSuccess)
+})
+  .catch(() => {
+    document.body.insertAdjacentHTML("beforeend", divError)
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    return reject();
+  }, 3000)
```
stanislav-topikha, line 31:
```suggestion
  setTimeout(reject, 3000)
```

stanislav-topikha (changes requested):
Check the comments ⬇️

liliia-livandovskaya (approved):
GJ!

PR #282, opened 2022-09-15
all done

Yevhen-Pavlovskyi:
https://yevhen-pavlovskyi.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
 // write your code here
+const logo = document.querySelector('.logo');
+
+function success() {
+  document.body.insertAdjacentHTML('beforeend', `
+   <div class="message">Promise was resolved!</div>`);
+}
+
+function error() {
+  setTimeout(() => {
+    document.body.insertAdjacentHTML('beforeend', `
+      <div class="error-message">Promise was rejected!</div>`);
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+    resolve(success());
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(error(), 3000);
+});
+
+promise1.finally(success);
+promise2.finally(error);
```
SanyaBratashchuk, line 29: image

SanyaBratashchuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+    resolve(success());
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(error(), 3000);
+});
+
+promise1.finally(success, error);
+promise2.finally(success, error);
```
SanyaBratashchuk, line 29: use then for happy flow and catch for handle errors, use finally only if some code must be executed anyway

SanyaBratashchuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #300, opened 2022-09-28
promise1 looks ok, but needs some changes for promise2

frontend112:
frontend112.github.io/js_promise_basic_DOM/


jstmpelowycz (approved):
nice!

PR #309, opened 2022-10-05
solution 1.0.0

mayuces:
- [DEMO LINK](https://mayuces.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,47 @@
 'use strict';
 
-// write your code here
+const body = document.getElementById('body');
+const logoButton = document.getElementById('logo');
+
+function createPromise1() {
+  const resolver = (logo) => {
+    logoButton.addEventListener('click', () => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,47 @@
...
+  const resolver = (resolve, cancel) => {
+    setTimeout(() => cancel(new Error()), 3000);
+  };
+
+  return new Promise(resolver);
+}
+
+const promise1 = createPromise1();
+const promise2 = createPromise2();
+
```
kamal-tufekcic, line 26:
```
const success = '<div class="message">Promise was resolved!</div>';
const error = '<div class="message error-message">Promise was rejected!</div>';
```

Wouldn't this be easier?

kamal-tufekcic (approved):
If it works it works, but there's so much needless code and variables, could be simplified a lot

PR #314, opened 2022-10-14
Solution

11Olya112:


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+logo.append(div);
+
+function createPromisse() {
+  const resolver = (complete) => {
+    logo.addEventListener('click', () => {
+      complete();
+    });
+  };
+
+  return new Promise(resolver);
```
Anastasiia-Svintsova, line 15: new Promise needs to have the resolve and reject methods

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+ordeal.then(onSuccess, onError);
+
+function onSuccess() {
+  div.className = 'message';
+  div.innerHTML = 'Promise was resolved!';
+}
+
+function onError() {
+  div.className = 'error-message';
```
Bohdan-Maksymiuk, line 28: looks like you need add one more className here 🙃

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-10-18 at 12 52 22

is it ok?)

Anastasiia-Svintsova (changes requested):
You need to create 2 different promises. It shouldn't look like that. The rejected promise shouldn't disappear
Animation

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+setTimeout(() => {
+  const promise2 = Promise.reject(new Error('Promise was rejected!'));
+
+  promise2
+    .catch(
+      div2.className = 'message error-message',
+      div2.innerHTML = 'Promise was rejected!'
+    );
+}, 3000);
```
qoa11a, line 28:
```suggestion
const promise1 = new Promise((resolve, reject) => {
  logo.addEventListener('click', () => {
    resolve();
  });
});

const promise2 = new Promise((resolve, reject) => {
  setTimeout(() => {
    reject();
  }, 3000);
});

div1.className = 'message';
div1.innerHTML = 'Promise was resolved!';

div2.className = 'message error-message';
div2.innerHTML = 'Promise was rejected!';

promise1.then(() => logo.append(div1));

promise2.catch(() => logo.append(div2));
```

qoa11a (approved):
Good job!
Here's another variant of solution of this task.

PR #317, opened 2022-10-19
firstPromises

fenderukr:
https://fenderukr.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,55 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', function() {
+    resolve();
+  });
+
+  document.body.addEventListener('click', function() {
+    reject(new Error('error'));
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error('error'));
+  }, 3000);
+});
+
+function myFunction() {
+  const divToAdd = document.createElement('div');
+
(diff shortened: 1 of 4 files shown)


polosanya (changes requested):
I think it shouldn't look like this 
image
Please fix it

fenderukr:
> I think it shouldn't look like this image Please fix it

hm, it doesn't look like this on my end
see the attached screenshot
Screenshot at Oct 20 17-26-24


what should i fix in my code?

src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error('error'));
+  }, 3000);
+});
+
+function myFunction() {
```
polosanya, line 21: Please give more informative names to your variables

src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', function() {
+    resolve();
+  });
+
+  document.body.addEventListener('click', function() {
+    reject(new Error('error'));
+  });
+});
```
polosanya, line 13: There is no such requirement in task description.
Promise1 should only be resolved on logo click.
You can leave it, but then please provide some styles to your error message

polosanya (changes requested):
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review 🙂 
Please check my comments. It would be great if you added some styles to your error message so they look something like the default message

PR #327, opened 2022-10-30
solution

JOWISSA:
[DEMO LINK](https://JOWISSA.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+
+const promise1 = new Promise(resolver1);
+
+function resolver1(complete, eror) {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    complete();
+  });
+}
+
+const promise2 = new Promise(resolver2);
+
+function resolver2(complete, eror) {
+  setTimeout(eror, 3000);
+};
+
+promise1.then(() => {
+  const message = document.createElement('div');
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+
+const promise1 = new Promise(resolver1);
+
+function resolver1(complete, eror) {
```
TarasHoliuk, line 7:
```suggestion
function callback1(resolve, reject) {
```
actually, the function (callback) that you passed to the new Promise has 2 arguments: resolve function and reject function (resolve and reject are default wide known names)
So to avoid confusion better to rename `resolver1`

TarasHoliuk: the same with resolver2

TarasHoliuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+
+const promise1 = new Promise(firstCallback);
+
+function firstCallback(complete, eror) {
```
marina-tilniak, line 7: bettre use resolve, reject for this function as in previous comment. This is more common

marina-tilniak (approved):

TarasHoliuk (approved):
GJ!
P.S. If the mentor approved PR but left a comment - it's not necessary to do additional changes. In this case, the comment is just FYI for the future :)

PR #331, opened 2022-11-02
Solution

anastasia-babko:
[DEMO LINK](https://anastasia-babko.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
```
DarkMistyRoom, line 5:
```suggestion
const promise1 = new Promise(resolve => {
```

DarkMistyRoom: You can live just resolve if you don't use reject

DarkMistyRoom (approved):

src/scripts/main.js
anastasia-babko: Got it, thanks.

PR #332, opened 2022-11-03
Solution

bokorzita:
[DEMO LINK](https://bokorzita.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -4,7 +4,7 @@ const logo = document.querySelector('.logo');
 const message = document.createElement('div');
 const body = document.body;
 
-new Promise((resolve, reject) => {
+new Promise((resolve) => {
   logo.addEventListener('click', () => {
     resolve();
   });


src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const message = document.createElement('div');
+const body = document.body;
+
+new Promise((resolve, reject) => {
```
jstmpelowycz, line 7: you can just omit `reject`
```suggestion
new Promise((resolve) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+}).then(() => {
+  message.classList.add('message');
+  message.innerHTML = `Promise was resolved!`;
+  body.appendChild(message);
+});
+
+new Promise((resolve, reject) => {
```
jstmpelowycz, line 17: here – as well (I believe linter would not complain)
```suggestion
new Promise((_, reject) => {
```

jstmpelowycz (approved):
great job!

PR #333, opened 2022-11-04
Promises were created

illia-kots:
[DEMO LINK](https://illia-kots.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,31 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+const messageResolved = document.createElement('div');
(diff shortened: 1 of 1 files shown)


loralevitska (approved):
You've done a nice job and everything on target. Cool!

A small recommendation for the future. You saw that there is a problem with the appearance of the jet, you can fix it a little to make it look better. In tests, it is very appreciated when you did a little more than what was required, because you care about the result and are ready for teamwork, and not just do what they say.

Before 
image
After -> if you study the dev-tools, you can see that the `message` class has the main classes for all messages, and the `error-message` class simply adjusts several settings 
image

illia-kots:
Ok. Thanks for recommendations! I will take it into account)

PR #336, opened 2022-11-07
solution

vitaliidox:
https://vitaliidox.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise(function(resolve) {
+  logo.addEventListener('click', () => {
+    document.body.innerHTML += `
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+});
+
+const promise2 = new Promise(function(resolve, reject) {
+  resolve(null);
+
+  setTimeout(() => {
+    document.body.innerHTML += `
+    <div class="error-message">
+      Promise was rejected!
+    </div>
```
vlad-demchuk, line 24: better not to use `innerHTML +=`
use `textContent`, `append, prepend` or `insertAdjacentHTML`

vlad-demchuk (changes requested):
call it using `then, catch...` to avoid these issues 
image

vlad-demchuk (approved):

PR #344, opened 2022-11-11
two promises created

iPanda177:
[DEMO LINK](https://iPanda177.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const root = document.querySelector('body');
+
+const promise1 = new Promise((resolve) => (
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+const logo = document.querySelector('.logo');
+const root = document.querySelector('body');
+
+const promise1 = new Promise((resolve) => (
+  logo.addEventListener('click', () => resolve())
+));
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => reject(new Error()), 3000);
+});
```
Anna-Tkachenko, line 12: will be better to create names according to the promise result: one of them should be the success one, other one should be the rejected one

Anna-Tkachenko (approved):

PR #355, opened 2022-11-29
promises created

miamarichka:
https://miamarichka.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('h1');
+
+  logo.addEventListener('click', () => {
+     resolve('Promise was resolved!');
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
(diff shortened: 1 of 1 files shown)


loralevitska (changes requested):
almost perfect) it could be better if you worry not only about the performance of the task, but also about the appearance of your work

improve UI 
(here I added `message` class)
image
(here is your solution)
image

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+    reject(new Error('Promise was rejected!'));
+  }, 3000);
+});
+
+promise1.then((resolvedText) => makeANotification('message', resolvedText));
+
+promise2.catch((errorText) => makeANotification('message error-message',
+  errorText));
+
+function makeANotification(classType, text) {
```
polosanya, line 22: It is a good tone to define a function before use. 
Referring to a function before declaration harms readability and maintainability.
Please move it to the top

polosanya (changes requested):
Almost perfect! Please check the comment to improve your code a bit

SanyaBratashchuk (approved):

PR #357, opened 2022-12-04
Develop

euFarafonov:


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', resolve);
+});
+
+promise1.then(onSuccess, onError);
```
witflash, line 9: It's better to handle `onError` in `catch` block because you may have any unexpected errors not only rejected promises.

witflash (approved):
Good job!

PR #361, opened 2022-12-09
Solution

Taras-1:
[DEMO LINK](https://Taras-1.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,38 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function createPromise() {
+  const resolver = (resolve, reject) => {
+    logo.addEventListener('click', () => {
+      resolve();
+    });
+  };
+
+  return new Promise(resolver);
+}
+
+const promise1 = createPromise();
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+promise1.then(() => {
+  success(success);
+
+  const promise2 = new Promise((resolve, reject) => {
+    setTimeout(() => reject(error), 3000);
+  });
+
+  return promise2;
+}, error)
+  .then(success, error);
```
nazarmatsevych, line 26: It's better to handle onError in catch block because you may have any unexpected errors not only rejected promises.

nazarmatsevych (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function createPromise() {
+  const resolver = (resolve, reject) => {
```
AlisaVlasova, line 6:
```suggestion
  const resolver = (resolve) => {
```

PR #372, opened 2022-12-25
The first promises were made

Andrii-Medintsev:
- [DEMO LINK](https://Andrii-Medintsev.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,34 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve) => {
+  document.querySelector('.logo').addEventListener('click', () => {
+    resolve('Promise was resolved!');
+  });
+});
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+  document.body.append(message);
+}
+
+promise1
+  .then(result => showNotification(result))
+  .catch(error => showNotification(error));
+
+promise2
+  .then(result => showNotification(result))
+  .catch(error => showNotification(error));
```
Anastasiia-Svintsova, line 34: You use the same logic in `then` and `catch` for both promises. Think about how not to repeat it.

Anastasiia-Svintsova (changes requested):

qoa11a (approved):

PR #386, opened 2023-01-07
Two promises

KateRiaz:
[demo](https://kateriaz.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+
+const resolver = (resolve, rejected) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+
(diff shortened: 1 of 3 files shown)


qoa11a (changes requested):
In your solution promise may be either resolved or rejected. But acording to the task one promise must be resolved on logo click and the other must be rejected after 3 seconds (so resolve and reject have to work both).

src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+});
+
+promise1
+  .then(
+    (result) => {
+      const divSuccess = document.createElement('div');
+
+      divSuccess.classList.add('message');
+      divSuccess.innerText = result;
+      body.append(divSuccess);
```
qoa11a, line 27: Create a function to show messages. DRY - don't repeat yourself.

qoa11a (changes requested):

qoa11a (approved):

PR #388, opened 2023-01-08
Solution

IVxVI:
- [DEMO LINK](https://ivxvi.github.io/js_promise_basic_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,52 @@
...
+promise2
+  .then(() => {
+    div2.className = 'message';
+    div2.innerHTML = successMessage;
+    document.body.append(div2);
+  })
+  .catch(() => {
+    div2.classList.add('message', 'error-message');
+    div2.innerHTML = errorMessage;
+    document.body.append(div2);
```
SanyaBratashchuk, line 51: create handle function and pass it to then and catch

SanyaBratashchuk (changes requested):

src/scripts/main.js
IVxVI: Nut sure I need both then and catch in both promises, so I tried to optimize it and created function that adds content

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+  setTimeout(() => {
+    reject(errorMessage);
+  }, 3000);
+});
+
+promise1
+  .then((result) => addElement(result, 'message'));
+
+promise2
+  .catch((result) => addElement(result, 'message error-message'));
```
agniya-i, line 29: you **must** have `then` and `catch` blocks for both promises, it doesn't matter if you know what they return

agniya-i (changes requested):

SanyaBratashchuk (approved):

PR #400, opened 2023-01-16
solution

divineretribution1:
[DEMO LINK](https://divineretribution1.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const body = document.querySelector('body');
+
+const resolver = (resolve, reject) => {
+  logo.addEventListener('click', () => resolve());
+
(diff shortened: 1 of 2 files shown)


qoa11a (changes requested):
There should be 2 promises: one that resolves on click and the other that rejects on setTimeout. In your case you have 2 identical promises that either resolve on click or reject on setTimeout.

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => resolve());
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => reject(new Error()), 3000);
+});
+
+const promises = (promise) => {
```
polosanya, line 14:
```suggestion
const handlePromises = (promise) => {
```

polosanya (approved):
Good job!

PR #401, opened 2023-01-16
solution

webdevw:
- [DEMO LINK](https://webdevw.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,38 @@
 'use strict';
 
-// write your code here
+const body = document.body;
+const message = document.createElement('div');
+
+function createPromise() {
+  const logo = document.querySelector('.logo');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+promise1
+  .then(result => {
+    message.className = 'message';
+    message.textContent = result;
+    body.append(message);
+  });
+
+promise2
+  .catch(error => {
+    message.className = 'error-message';
```
loralevitska, line 35:
```suggestion
    message.className = 'message error-message';
```

loralevitska (changes requested):
Almost done) Now work only 1 promise. Or first, or second. Try to fix this. They should work together

qoa11a (approved):

PR #402, opened 2023-01-16
Develop

tarasavage:
[DEMO LINK](https://tarasavage.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -15,11 +15,11 @@ const promise2 = new Promise((resolve, reject) => {
 });
 
 function promiseHandler(className, message) {
-  document.body.insertAdjacentHTML('beforeend', `
-    <div class=${className}>
-      Promise was ${message}!
-    </div>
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error());
+  }, 3000);
+});
+
+function promiseHandler(className, message) {
+  document.body.insertAdjacentHTML('beforeend', `
+    <div class=${className}>
```
witflash, line 19: Could you use `const messageElement = document.createElement('div');` approach instead of raw text?
And add the class name and text node to the messageElement.

witflash (changes requested):
It may be as you have done but could you fix it with my comment?

IvanFesenko (approved):

PR #406, opened 2023-01-21
solution

halynarydasova:
- [DEMO LINK](https://halynarydasova.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,40 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function onSuccess() {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+});
+
+promise1.then(
+  value => onSuccess(),
+  error => onError(error)
+);
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    // eslint-disable-next-line prefer-promise-reject-errors
```
polosanya, line 32: Don't disable the eslint, but fix its problem.
Try googling the error text to understand it

polosanya (approved):

PR #419, opened 2023-02-01
js_promise_basic_DOM solution

ysrckr:
[DEMO LINK](https://ysrckr.github.io/js_promise_basic_DOM/)


meforoma (approved):
✅

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+  }, 3000);
+});
+
+promise1
+  .then((data) => {
+    const successDiv = document.createElement('div');
+
+    successDiv.classList.add('message');
+    successDiv.innerText = data;
+    body.appendChild(successDiv);
```
meforoma, line 24: i believe you can DRY the code drastically

meforoma (changes requested):
let's DRY those try and catch code blocks

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+};
+
+const createErrorMessage = (err) => {
+  const body = document.querySelector('body');
+  const errorDiv = document.createElement('div');
+
+  errorDiv.classList.add('error-message');
+  errorDiv.innerText = err;
+  body.appendChild(errorDiv);
+};
```
meforoma, line 33: i'd DRY it even more :)
```suggestion
const createSuccessMessage = (
  type: 'data' | 'error', data
) => {
  const body = document.querySelector('body');
  const div = document.createElement('div');
  const classname = type === 'message'
    ? 'message'
    : 'error-message';

  div.classList.add(classname);
  div.innerText = data;
  body.appendChild(div);
};
```

meforoma (approved):
✅

PR #425, opened 2023-02-07
Solution

olena-hapon:
[DEMO LINK](https://olena-hapon.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -2,19 +2,10 @@
 
 const logo = document.querySelector('.logo');
 
-const promise1 = new Promise((resolve, reject) => {
+const promise1 = new Promise((resolve) => {
   logo.addEventListener('click', () => {
     resolve('Promise was resolved!');
   });
-});
-
-promise1.then(value => {
-  const div = document.createElement('div');
-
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
```
AlisaVlasova, line 5:
```suggestion
const promise1 = new Promise((resolve) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+});
+
+promise1.then(value => {
+  const div = document.createElement('div');
+
+  div.innerText = value;
+  div.className = 'message';
+
+  document.body.append(div);
+});
```
AlisaVlasova, line 18: it would be great if you move message creation logic into function and reuse it

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #426, opened 2023-02-09
Develop

anastasia-radchencko:
https://anastasia-radchencko.github.io/js_promise_basic_DOM/


Volodymyr-Melnyk-Fron10der (approved):
Nice job, I like it

PR #433, opened 2023-02-16
solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/js_promise_basic_DOM/


jaxxxon78:
Your code actually isn't working as it is prescribed.

PR #437, opened 2023-02-20
added

vovakov123:
- [DEMO LINK](https://vovakov123.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function complete(result) {
+  const div = document.createElement('div');
+
+  div.innerHTML = 'Promise was resolved!';
+  div.className = 'message';
+  document.body.appendChild(div);
+}
+
+function error(data) {
+  const div = document.createElement('div');
+
+  div.innerHTML = 'Promise was rejected!';
+  div.className = 'error-message';
+  document.body.appendChild(div);
+}
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(error);
+  }, 3000);
+});
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function complete(result) {
+  const div = document.createElement('div');
+
+  div.innerHTML = 'Promise was resolved!';
```
sTorba24, line 8:
```suggestion
  div.textContent= 'Promise was resolved!';
```

sTorba24: Use right methods. You are passing just text, it's not html, so you don't need to use `innerHTML` :)

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(error);
+  }, 3000);
+});
+
+promise1
+  .then(complete)
+  .catch(error);
```
sTorba24, line 35: not needed .catch here

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+
+promise2
+  .then(complete)
+  .catch(error);
```
sTorba24, line 39: remove .then here

sTorba24 (changes requested):
Good work! Let's improve it a little :)

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+
+  div.textContent = 'Promise was rejected!';
+  div.className = 'error-message';
+  document.body.appendChild(div);
+}
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
```
anstsot, line 24: you can write it more simply if you want  
```suggestion
  logo.addEventListener('click', resolve);
```

anstsot (approved):
Great job, approved!

PR #440, opened 2023-02-21
add 2 promises

NikitaNevv:
https://NikitaNevv.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
+// debugger
 
-// write your code here
+const logo = document.querySelector('.logo');
+const body = document.querySelector('body');
+
+function success() {
+  function positveResult(resolve) {
+    logo.addEventListener('click', () => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
 'use strict';
+// debugger
```
TarasHoliuk, line 2: do not leave comments

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+    logo.addEventListener('click', () => {
+      resolve('Promise was resolved!');
+    });
+  }
+
+  return new Promise(positveResult);
+}
+
+function fail() {
+  const failed = (resolved, error) => {
```
TarasHoliuk, line 18: the first time you create an internal function via function declaration and the second time via function expression
be consistent in such things

Just FYI

TarasHoliuk (approved):
Approved :heavy_check_mark:

PR #443, opened 2023-02-23
 js_promise_basic_DOM task done

EhorSaprykin:
- [DEMO LINK](https://ehorsaprykin.github.io/js_promise_basic_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+function showNotification(message, cls) {
+  const div = document.createElement('div');
+  const body = document.querySelector('body');
+
+  div.classList.add(cls);
+  div.textContent = message;
+  body.append(div);
+};
+
+const promise1 = Promise.resolve(() => {});
```
qoa11a, line 14: you resolve your promise right away, while it should be resolved on click on logo

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+const promise2 = Promise.reject(new Error());
```
qoa11a, line 15: this promise should be rejected after 3 seconds and you reject it right away. You just show message in 3 seconds, but the promise is rejected right away

qoa11a (changes requested):

src/scripts/main.js
EhorSaprykin: @qoa11a done

src/scripts/main.js
EhorSaprykin: @qoa11a done

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve('Promise was resolved!');
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() =>
+    reject(new Error('Promise was rejected!')), 3000);
```
Volodymyr-Melnyk-Fron10der, line 22: I would move time into constant

Volodymyr-Melnyk-Fron10der (approved):
Well done✅ Take a look on one recommendation for future

PR #444, opened 2023-02-24
Js promise basic task was done

SamirMamedov1010:
- [DEMO LINK](https://SamirMamedov1010.github.io/js_promise_basic_DOM/)


ViktorHavryliuk (changes requested):
Promise 1 should not be rejected, it can only be resolved even if it takes more than 3 seconds. Promise 2 should be relected in 3 seconds regardless of wether promise 1 was resolved or not.

mykhalenych (approved):

PR #445, opened 2023-02-24
Solution

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+const mateLogo = document.querySelector('.logo');
+
+function createMessage(message, type = '') {
+  const div = document.createElement('div');
+
+  div.innerHTML = `<div class="message ${type}">${message}</div>`;
+  document.body.append(div);
+}
+
+const promise1 = new Promise((resolve) => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+}
+
+const promise1 = new Promise((resolve) => {
+  mateLogo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => reject(new Error('Promise was rejected!')), 3000);
```
loralevitska, line 19: Overall not bad, but with the text it is not what is expected in the task. Doing a little better and more than described in the task is very cool, but not with content. Text and logic must always be reproduced as described in the task. But the fact that you improved the visual is great
```suggestion
  setTimeout(() => reject('Promise was rejected!'), 3000);
```

image

loralevitska (approved):
👍

src/scripts/main.js
pushkarskiyrodion: Thanks a lot. I'll do my best )

PR #451, opened 2023-03-01
Solution

Ukrainiane-panda:
[DEMO LINK](https://Ukrainiane-panda.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
+const promise1 = new Promise((resolve) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', e => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(Error);
+  }, 3000);
+});
+
+promise1.then(() => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+});
+
+promise1.then(() => {
+  const message = document.createElement('div');
+
+  message.classList.add('message');
+
+  message.textContent = 'Promise was resolved!';
+
+  document.body.append(message);
```
DarkMistyRoom, line 24: move `message `logic to a separate function and reuse it to avoid duplicating

DarkMistyRoom (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(Error);
+  }, 3000);
+});
+
+const createElement = (classForDiv, text) => {
```
DarkMistyRoom, line 17:
```suggestion
const createMessage = (classForDiv, text) => {
```

DarkMistyRoom (approved):

PR #453, opened 2023-03-01
Solution

stanislavcodes:
[DEMO LINK](https://stanislavcodes.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
 
-// write your code here
+function createMessage(classname, message) {
+  const messageDiv = document.createElement('div');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+}
+
+function onError(error) {
+  document.body.appendChild(
+    createMessage('error-message', error.message)
+  );
+}
+
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
```
illiarerun, line 25: getElementsByClassName also would work, but it'll return `NodeList`

illiarerun (approved):

PR #464, opened 2023-03-08
[SOLUTION] Added two promise

grabynsky:


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+});
+
+promise1.then(() => {
+  body.insertAdjacentHTML('beforeend',
+    `<div class="message">
+      Promise was resolved!
+    </div>`);
+});
+
+promise2.catch(() => {
```
vlad-demchuk, line 24: image
check this requirement point

vlad-demchuk (changes requested):

IvanFesenko (approved):
Well done

PR #469, opened 2023-03-10
first promise commit

haduigon:
https://haduigon.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,47 @@
 'use strict';
 
 // write your code here
+
+const logo = document.querySelector('.logo');
+const bd = document.querySelector('body');
+
+logo.addEventListener('click', () => {
+  const success = () => new Promise((resolve, reject) => {
+    function resolveHandler() {
+      const newDiv = document.createElement('div');
+
+      newDiv.className = 'message';
+      newDiv.innerText = 'Promise was resolved!';
+      bd.appendChild(newDiv);
+
+      setTimeout(() => {
+        newDiv.remove();
+      }, 3000);
+      resolve('ok');
+    }
+
+    return resolveHandler();
+  });
+
+  const fail = () => new Promise((resolve, reject) => {
+    function rejectHandler() {
+      const newDivError = document.createElement('div');
+
+      newDivError.className = 'error-message';
+      newDivError.innerText = 'Promise was rejected!';
+      bd.appendChild(newDivError);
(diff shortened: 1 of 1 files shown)


etojeDenys (changes requested):
you need to write three promises and in these promises add addEventListener

src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
 'use strict';
 
 // write your code here
+
+const logo = document.querySelector('.logo');
```
tetiana-ivanchenko, line 5:
```suggestion

const logo = document.querySelector('.logo');
```

src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
...
+const bd = document.querySelector('body');
+
+const rejected = (resolve, reject) => {
+
+
+  function rejectHandler() {
```
tetiana-ivanchenko, line 11:
```suggestion
const rejected = (resolve, reject) => {
  function rejectHandler() {
```

src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
...
+    const newDivError = document.createElement('div');
+    setTimeout(() => {
+
+      newDivError.className = 'error-message';
+      newDivError.innerText = 'Promise was rejected!';
+      bd.appendChild(newDivError);
+      reject(new Error('not ok'));
```
tetiana-ivanchenko, line 18: it should be in the catch function

src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
...
+    function resolveHandler() {
+      const newDiv = document.createElement('div');
+
+      newDiv.className = 'message';
+      newDiv.innerText = 'Promise was resolved!';
+      bd.appendChild(newDiv);
+
+      setTimeout(() => {
+        newDiv.remove();
+      }, 3000);
```
tetiana-ivanchenko, line 42: it should be in the then function

tetiana-ivanchenko (changes requested):

src/scripts/main.js
haduigon: have put the functions inside then & catch

sTorba24 (changes requested):
first promise should be resolved, second rejected
Not needed to remove error message, it's not asked in task description

Viktor-Kost (approved):

PR #478, opened 2023-03-19
solution :)

katerynashylina:
- [DEMO LINK](https://katerynashylina.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(reject, 3000);
+});
+
+promise1
+  .then(onSuccess)
+  .catch(onError);
+
+promise2
+  .then(onSuccess)
```
TarasHoliuk, line 20: it can't be resolved, so you don't have to add `then` here - FYI

TarasHoliuk (approved):

PR #486, opened 2023-03-24
Solution

fisherok3452:
[DEMO LINK](https://fisherok3452.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,26 @@
+/* eslint-disable no-unused-vars */
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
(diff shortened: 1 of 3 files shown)


OlEzhA0 (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
+/* eslint-disable no-unused-vars */
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
```
OlEzhA0, line 6: next time, would be better to resolve message from promise, and handle it 
promise1.then().catch()

PR #492, opened 2023-03-29
js_promise_basic_DOM

JulyaPetrovskaya:
https://JulyaPetrovskaya.github.io/js_promise_basic_DOM

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -5,6 +5,7 @@ const logoButton = document.querySelector('.logo');
 const promise1 = new Promise(resolve => {
   logoButton.addEventListener('click', () => {
     const div1 = document.createElement('div');
+
     div1.classList.add('message');
     div1.textContent = 'Promise was resolved!';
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
-// write your code here
+const logoButton = document.querySelector('.logo');
+
+const promise1 = new Promise(resolve => {
+  logoButton.addEventListener('click', () => {
+    const div1 = document.createElement('div');
+
+    div1.classList.add('message');
+    div1.textContent = 'Promise was resolved!';
+    document.body.appendChild(div1);
```
IvanFesenko, line 11: Move it to `then` of the first promise.
Here you should call `resolve`
```suggestion
  resolve();
```

mykhalenych (changes requested):
don't see any changes after previous review, check comments

mykhalenych (changes requested):
deploy link isn't working

PR #494, opened 2023-03-29
task done

andriy-litkovets:
[DEMO LINK](https://andriy-litkovets.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+    3000
+  );
+});
+
+promise1
+  .then()
+  .catch((err) => alert(err));
+
+promise2
+  .then((err) => alert(err))
```
polosanya, line 24: It's not an error, as then method executes only on success)
It's just not what we need, but it is certainly not an error

polosanya (approved):
Good job!

PR #495, opened 2023-03-30
js_promise_basic_DOM

JulyaPetrovskaya:
- [DEMO LINK](https://JulyaPetrovskaya.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const logoButton = document.querySelector('.logo');
+
+const promise1 = new Promise(resolve => {
+  logoButton.addEventListener('click', () => {
+    resolve();
+  });
+});
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+  document.body.appendChild(div1);
+});
+
+promise2.catch((message) => {
+  const div2 = document.createElement('div');
+
+  div2.classList.add('error-message');
+  div2.classList.add('message');
+  div2.textContent = message;
+  document.body.appendChild(div2);
```
OlEzhA0, line 31: can you move this logic to some function like `createMessage` and use it here?

OlEzhA0 (changes requested):

VitaliyBondarenko1982 (approved):

PR #498, opened 2023-03-31
solution

Ilya-Osoben:
[DEMO LINK](https://Ilya-Osoben.github.io/js_promise_basic_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    setTimeout(() => reject(new Error('error')), 3000);
+  });
```
etojeDenys, line 22:
```suggestion
    setTimeout(reject, 3000);
```

just call reject after 3 seconds

etojeDenys (changes requested):

PR #503, opened 2023-04-03
solution

DanielSorokowski:
- [DEMO LINK](https://DanielSorokowski.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,43 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(Error);
+  }, 3000);
+});
+
+const error = () => {
+  const div = document.createElement('div');
+
+  div.classList.add('error-message');
+  div.innerText = `Promise was rejected!`;
+
+  document.body.appendChild(div);
+};
+
+const success = () => {
+  const div = document.createElement('div');
+
+  div.classList.add('message');
+  div.innerText = `Promise was resolved!`;
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
```
solaryasha, line 5:
```suggestion
const waitForClick = new Promise((resolve) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
```
solaryasha, line 11:
```suggestion
const waitForDelay = new Promise((_, reject) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+};
+
+const success = () => {
+  const div = document.createElement('div');
+
+  div.classList.add('message');
+  div.innerText = `Promise was resolved!`;
+
+  document.body.appendChild(div);
+};
```
solaryasha, line 33: This function has a lot in common. They only differs with message and inner text

solaryasha: Let's write an `addMessage` function that takes two args

solaryasha (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const waitForClick = new Promise((resolve, reject) => {
```
solaryasha, line 5: You don't use `reject` function in the promise
```suggestion
const waitForClick = new Promise((resolve) => {
```

solaryasha (approved):

PR #507, opened 2023-04-06
Develop

ilona-romaniuk:
https://ilona-romaniuk.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+
+function addThenCatchToPromise(promise) {
+  return promise
+    .then(() => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    const error = true;
+
+    reject(error);
```
Volodymyr-Melnyk-Fron10der, line 30:
```suggestion
    reject(true);
```

Volodymyr-Melnyk-Fron10der: useless variable

Volodymyr-Melnyk-Fron10der (changes requested):
one fix I see

Anastasiia-Svintsova (approved):

PR #509, opened 2023-04-10
solution

popeye88:


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+    'beforeend',
+    `
+    <div class="message">
+      Promise was resolved!
+    </div>
+  `
+  );
+}
+
+function isRejected() {
```
polosanya, line 26:
```suggestion
function handleError() {
```
Actually, these two functions are similar and can be combined into one function

polosanya (approved):
Good job! It is better to add only content of src folder and readme file to your commit next time. Also, don't forget to add demo and test link to your Pull Request description
image
image

popeye88:
Thanks for reminder. My fault)

PR #511, opened 2023-04-11
solution

MikiOOOO:
- [DEMO LINK](https://MikiOOOO.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
 // write your code here
+const body = document.querySelector('body');
+
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    resolve('Promise was resolved!');
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


danielpopek94 (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+});
+
+promise2.then(() => {
+  return 'success';
+}).catch(information => {
+  const div = document.createElement('div');
+
+  div.classList.add('error-message');
+  div.textContent = information;
+  body.appendChild(div);
```
d-wasilewski, line 35: These functions has a lot in common. They only differs with message and text. Let's write an addMessage function that takes two args `message` and `type`

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
 'use strict';
 
 // write your code here
+const body = document.querySelector('body');
+
+const promise1 = new Promise((resolve, reject) => {
```
d-wasilewski, line 6: You don't use reject here
```suggestion
const promise1 = new Promise((resolve) => {
```

d-wasilewski (changes requested):

PR #512, opened 2023-04-11
Develop

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
+
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error('something bad happened'));
+  }, 3000);
+});
+
+promise1.then(() => {
+  const message = document.createElement('div');
+
+  message.classList.add('message');
+  message.textContent = 'Promise was resolved!';
+  document.querySelector('body').append(message);
+});
+
+promise2.catch(() => {
+  const message = document.createElement('div');
+
+  message.classList.add('error-message');
+  message.textContent = 'Promise was rejected!';
+  document.querySelector('body').append(message);
+});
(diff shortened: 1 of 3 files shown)


mykhalenych (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+const promise1 = new Promise(resolve => {
+  logo.addEventListener('click', () => {
+    resolve(document.body.append(successContainer));
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(document.body.append(errorContainer));
```
etojeDenys, line 22: Instead of rejecting with 'document.body.append(errorContainer)', just reject and move the 'document.body.append()' operation to the '.catch()' block.

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+successContainer.classList.add('message');
+successContainer.textContent = 'Promise was resolved!';
+
+errorContainer.classList.add('error-message');
+errorContainer.classList.add('message');
+errorContainer.textContent = 'Promise was rejected!';
+
+const promise1 = new Promise(resolve => {
+  logo.addEventListener('click', () => {
+    resolve(document.body.append(successContainer));
```
etojeDenys, line 16: Instead of resolving with 'document.body.append(successContainer)', just resolve and move the 'document.body.append()' operation to the '.then()' block.

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(document.body.append(errorContainer));
+  }, 3000);
+}
+);
+
+promise1.then();
+promise2.then();
```
etojeDenys, line 28: Use '.catch()' to append 'errorContainer' to the body after 'promise2' rejects.

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(document.body.append(errorContainer));
+  }, 3000);
+}
+);
+
+promise1.then();
```
etojeDenys, line 27: Use '.then()' to append 'successContainer' to the body after 'promise1' resolves.

etojeDenys (changes requested):

PR #519, opened 2023-04-16
solution

danielpopek94:
[DEMO LINK](https://danielpopek94.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => resolve());
+});
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+  document.body.appendChild(resolveMessage);
+};
+
+const onError = () => {
+  const rejectMessage = document.createElement('div');
+
+  rejectMessage.classList.add('error-message');
+  rejectMessage.textContent = 'Promise was rejected!';
+  document.body.appendChild(rejectMessage);
+};
```
DanielSorokowski, line 27: you can create one function that takes 2 args and call it with diff values

DanielSorokowski (changes requested):

d-wasilewski (changes requested):

PR #520, opened 2023-04-17
Solution

Vaskovics:
<a href="https://vaskovics.github.io/js_promise_basic_DOM/">Demo link</a>

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,49 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+promise1
+  .then(() => {
+    const div = document.createElement('div');
+
+    div.className = 'message';
+    div.innerText = 'Promise was resolved!';
+    document.body.append(div);
+  })
+  .catch(() => {
+    const div = document.createElement('div');
+
+    div.className = 'error-message';
+    div.innerText = 'Promise was rejected!';
+    document.body.append(div);
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    const er = 'Promise was rejected!';
+
+    reject(er);
+  }, 3000);
+});
+
+promise2
+  .then(() => {
+    const div = document.createElement('div');
+
+    div.className = 'message';
+    div.innerText = 'Promise was resolved!';
+    document.body.append(div);
+  })
+  .catch((er) => {
+    const div = document.createElement('div');
+
+    div.className = 'error-message';
+    div.innerText = er;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+    const div = document.createElement('div');
+
+    div.className = 'error-message';
+    div.innerText = 'Promise was rejected!';
+    document.body.append(div);
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    const er = 'Promise was rejected!';
```
mykhalenych, line 29: To make the code more readable, consider changing the variable name `er` to something more descriptive like `errorMessage`.

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+
+promise2
+  .then(() => {
+    const div = document.createElement('div');
+
+    div.className = 'message';
+    div.innerText = 'Promise was resolved!';
+    document.body.append(div);
+  })
+  .catch((er) => {
```
mykhalenych, line 43:
```suggestion
  .catch((error) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+  setTimeout(() => {
+    const er = 'Promise was rejected!';
+
+    reject(er);
+  }, 3000);
+});
+
+promise2
+  .then(() => {
+    const div = document.createElement('div');
```
mykhalenych, line 37: better to create function handler for success and error and use this function in first case
`.then(onSuccess)`
in second `.catch`

mykhalenych (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+    reject(mesage);
+  }, 3000);
+});
+
+promise2
+  .then(() => {
+    createMessage('Promise was resolved!', 'message');
+  })
+  .catch((message) => {
+    createMessage('Promise was rejected!', 'error-message');
```
polosanya, line 40: You should use the `message` parameter from the catch block instead of hardcoding the string 'Promise was rejected!'. Example: `createMessage(message, 'error-message');`

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+  .then(() => {
+    createMessage('Promise was resolved!', 'message');
+  })
+  .catch(() => {
+    createMessage('Promise was rejected!', 'error-message');
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    const mesage = 'Promise was rejected!';
```
polosanya, line 29: There is a typo in the variable name `mesage`. It should be `message`. Please correct it.

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments

PR #522, opened 2023-04-17
promise

DmytroHorshunov:
https://DmytroHorshunov.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+promise1
+  .then(() => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+promise1
+  .then(() => {
+    document.body.insertAdjacentHTML('beforeend', `
+    <div class="message">Promise was resolved!</div>
+    `);
```
anstsot, line 15: you have written the same code several times. please create separate function `createNotification` with params and use it everywhere

anstsot (changes requested):
It’s almost done! Let’s improve a little

src/scripts/main.js
DmytroHorshunov: i think, that i right understood

IvanFesenko (approved):
Well done

PR #523, opened 2023-04-17
Develop

Dka888:
[DEMO LINK](https://Dka888.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -13,7 +13,7 @@ promise1.then(resolution => {
   div.classList.add('message');
   div.textContent = resolution;
   document.body.append(div);
-}).catch(Error('error'));
+}).catch(Error('error!'));
 
 const promise2 = new Promise(function(resolve, reject) {
   setTimeout(() => reject(new Error(`Promise was rejected!`)), 3000);
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+let div;
+const logo = document.querySelector('logo');
+const promise1 = new Promise(function(resolve, reject) {
+  logo.addEventListener('click', () => {
+    resolve(`Promise was resolved!`);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+let div;
+const logo = document.querySelector('.logo');
+const promise1 = new Promise(function(resolve, reject) {
```
d-wasilewski, line 5:
```suggestion
const waitForClick = new Promise(function(resolve, reject) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+});
+
+promise1.then(resolution => {
+  div = document.createElement('div');
+  div.classList.add('message');
+  div.textContent = resolution;
+  document.body.append(div);
+}).catch(Error('error!'));
+
+const promise2 = new Promise(function(resolve, reject) {
```
d-wasilewski, line 18:
```suggestion
const waitForDelay = new Promise(function(resolve, reject) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+});
+
+promise2.then(() => {
+  return 'Succes';
+})
+  .catch(rejection => {
+    div = document.createElement('div');
+    div.classList.add('error-message');
+    div.textContent = rejection;
+    document.body.append(div);
```
d-wasilewski, line 29: These functions has a lot in common. They only differs with message and text. Let's write an addMessage function that takes two args `message` and `type`

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+let div;
```
d-wasilewski, line 3:
```suggestion
```

d-wasilewski (changes requested):

PR #529, opened 2023-04-19
Solution

mykolahuryn:
- [DEMO LINK](https://mykolahuryn.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
-'use strict';
+'use strict'
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  document.querySelector('.logo').addEventListener('click', () => {
+    resolve('Promise was resolved!');
+  });
+});
+
+promise1.then((result) => {
+  const message = document.createElement('div');
+
+  message.classList.add('message');
+  message.textContent = result;
+  document.body.appendChild(message);
+
+  const promise2 = new Promise((resolve, reject) => {
+    setTimeout(() => {
+      resolve('Promise was rejected!');
+    }, 3000);
+  });
+
+  return promise2;
+}).then((result) => {
+  const message = document.createElement('div');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+promise1.then((result) => {
+  const message = document.createElement('div');
+
+  message.classList.add('message');
+  message.textContent = result;
+  document.body.appendChild(message);
+
+  const promise2 = new Promise((resolve, reject) => {
+    setTimeout(() => {
+      resolve('Promise was rejected!');
```
tetiana-ivanchenko, line 18: The provided text should be 'Promise was rejected!', but you have resolved the promise with this text. Please change `resolve` to `reject` to correctly reject the promise. 🚀

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
-'use strict';
+'use strict'
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
```
tetiana-ivanchenko, line 3: It looks like both promises should be declared at the beginning, but `promise2` is declared inside the `then` of `promise1`. Please move the declaration of `promise2` to the top, and make sure it's rejected instead of resolved. 😊

tetiana-ivanchenko (changes requested):
image

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+const error = () => {
+  body.insertAdjacentHTML('beforeend', 
+  `<div class="message error-message">Promise was rejected!</div>`
+  )
+}
+
+promise1.then(success, error);
+promise2.then(success, error);
\ No newline at end of file
```
polosanya, line 28: Add a newline at the end of the file for better readability and to follow best practices.
Also, better catch reject using catch() method
image

polosanya (approved):
Good job!

PR #535, opened 2023-04-24
Solution

dudar-y:
[DEMO LINK](https://dudar-y.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+const page = document.querySelector('body');
+
+const promise1 = new Promise((resolve) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    resolve(compleat);
+  });
+});
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+
+const compleat = () => {
+  page.insertAdjacentHTML('beforeend',
+    `<div class = 'message'>Promise was resolved!</div>`);
+};
+
+const error = () => {
+  page.insertAdjacentHTML('beforeend',
+    `<div class = 'message error-message'>Promise was rejected!</div>`);
+};
```
DarkMistyRoom, line 37: would be better to create `showMesssage `function wich will take message and show it

DarkMistyRoom: then you can resolve/reject promises with messages and run this function in `then`. It will unify showing message logic

DarkMistyRoom (changes requested):
Good job! Let's improve the code a bit

loralevitska (approved):
🔥

PR #536, opened 2023-04-24
1st version

Grechana:
https://Grechana.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const div = document.createElement('div');
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve(12);
+  });
+});
+
+promise1
+  .then(() => {
+    body.append(div);
+    div.classList = 'message';
+    div.textContent = 'Promise was resolved!';
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error('Error'));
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const div = document.createElement('div');
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve(12);
```
polosanya, line 9: What is 12?

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+  });
+});
+
+promise1
+  .then(() => {
+    body.append(div);
+    div.classList = 'message';
+    div.textContent = 'Promise was resolved!';
```
polosanya, line 17: It would be great to create a separate function for showing a message as it's almost the same for success and reject.
Also, you can get an argument of resolve\reject as then\catch parameter

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+  .catch(() => {
+    body.append(div);
+    div.classList = 'error-message';
+    div.textContent = 'Promise was rejected!';
+  });
+
+// logo.addEventListener('click', (e) => {
+//     console.log(e.tagret);
+//  console.log('11');
+//         });
```
polosanya, line 36: Don't forget to remove the comments

polosanya (changes requested):
Good job! Let's improve your solution a bit)

OleziO (approved):

PR #551, opened 2023-05-07
first

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
-// write your code here
+function createPromise() {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      // eslint-disable-next-line prefer-promise-reject-errors
+      reject();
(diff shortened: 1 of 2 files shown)


SanyaBratashchuk (changes requested):
promise2 is not working

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
 'use strict';
 
-// write your code here
+function createPromise() {
```
DarkMistyRoom, line 3: There is no reason to create this function, as we have two different promises and this logic doesn't repeat. Usually, it is good to separate logic but in this case, it makes no sense, because we use it only once

DarkMistyRoom (changes requested):
It works good, but let's improve the code style a bit

anstsot (approved):
Great job, approved!

PR #561, opened 2023-05-15
solution

AnastasiiaVorobets:
- [DEMO LINK](https://AnastasiiaVorobets.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve) => {
+  logo.addEventListener('click', () => {
+    resolve();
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(addError);
+  }, 3000);
+});
+
+function addSuccess() {
+  document.body.insertAdjacentHTML('beforeend', `
+    <div class="message">
+      Promise was resolved!
```
vlad-demchuk, line 20: you can pass your message to resolve, reject methods and take them in then/catch

you can make your functions more reusable in this way

vlad-demchuk (approved):

PR #565, opened 2023-05-18
promise-basic-DOM first try

RufatRM:
https://RufatRM.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
 // write your code here
+const logo = document.querySelector('.logo');
+const succesDiv = document.createElement('div');
+const errorDiv = document.createElement('div');
+
+succesDiv.classList.add('message');
+errorDiv.classList.add('message', 'error-message');
+
+const succesMessage = document.createTextNode('Promise was resolved!');
+const errorMessage = document.createTextNode('Promise was rejected!');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = Promise.reject;
+
+promise1
+  .then(() => {
+    succesDiv.appendChild(succesMessage);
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+const succesMessage = document.createTextNode('Promise was resolved!');
+const errorMessage = document.createTextNode('Promise was rejected!');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = Promise.reject;
```
qoa11a, line 20: You should reject your promise in 3 seconds. You need to use syntax like above `new Promise((resolve, reject) => {})` but just use `setTimeout` inside and reject only after 3 seconds has passed.

qoa11a (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
 
 // write your code here
+const logo = document.querySelector('.logo');
+const succesDiv = document.createElement('div');
+const errorDiv = document.createElement('div');
+
+succesDiv.classList.add('message');
+errorDiv.classList.add('message', 'error-message');
+
+const succesMessage = document.createTextNode('Promise was resolved!');
```
Anastasiia-Svintsova, line 11: Typo in variable name, change 'succesMessage' to 'successMessage'.

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
 'use strict';
 
 // write your code here
+const logo = document.querySelector('.logo');
+const succesDiv = document.createElement('div');
```
Anastasiia-Svintsova, line 5:
```suggestion
const successContainer = document.createElement('div');
```

Anastasiia-Svintsova (approved):

PR #577, opened 2023-05-24
js_promise-basic-DOM

vintwp:
https://vintwp.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    resolve('Promise was resolved!');
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error('Promise was rejected!'));
+  }, 3000);
+});
+
+promise1
+  .then(result => {
+    const message = createMessage(result);
+
+    document.body.append(message);
+  });
+
+promise2
+  .catch(result => {
+    const message = createMessage(result);
+
+    message.classList.add('error-message');
+    document.body.append(message);
+  });
+
+function createMessage(innerMessage) {
+  const message = document.createElement('div');
+
+  message.classList.add('message');
+  message.innerHTML = `<p>${innerMessage}</p>`;
+
+  return message;
+}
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    resolve('Promise was resolved!');
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error('Promise was rejected!'));
```
vlad-demchuk, line 13: Use 'reject('Promise was rejected!')' instead of 'reject(new Error('Promise was rejected!'))' to avoid creating an Error object.

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+promise1
+  .then(result => {
+    const message = createMessage(result);
+
+    document.body.append(message);
+  });
+
+promise2
+  .catch(result => {
+    const message = createMessage(result);
```
vlad-demchuk, line 26: Replace 'result' with 'error' for better readability in the catch block, as it represents an error.

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  setTimeout(() => {
+    reject(new Error('Promise was rejected!'));
+  }, 3000);
+});
+
+promise1
+  .then(result => {
+    const message = createMessage(result);
+
+    document.body.append(message);
```
vlad-demchuk, line 21: image

vlad-demchuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
 'use strict';
 
-// write your code here
+const errorMessage = () => {
```
polosanya, line 3:
```suggestion
const showErrorMessage = () => {
```
Functions should be named according to these rules
image

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  const message = document.createElement('div');
+
+  message.classList.add('message', 'error-message');
+  message.innerHTML = 'Promise was rejected!';
+  document.body.append(message);
+};
+
+const succesMessage = () => {
```
polosanya, line 11: Same here. Also, functions for showing success and error messages are almost the same. It would be great to combine them into one function

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments

PR #598, opened 2023-06-08
solution

natalia-maryshko:
[DEMO LINK](https://natalia-maryshko.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+const resolve = () => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    const message = document.createElement('div');
+
+    message.classList.add('message');
+    message.textContent = 'Promise was resolved!';
+    document.body.append(message);
+  });
+};
+
+const reject = () => {
+  setTimeout(() => {
+    const message = document.createElement('div');
+
+    message.classList.add('message', 'error-message');
+    message.textContent = 'Promise was rejected!';
+    document.body.append(message);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+const resolve = () => {
```
polosanya, line 3: Create a promise with the `resolve` function as the executor. The `resolve` function should not add an event listener, but instead directly resolve the promise.

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+  logo.addEventListener('click', () => {
+    const message = document.createElement('div');
+
+    message.classList.add('message');
+    message.textContent = 'Promise was resolved!';
+    document.body.append(message);
+  });
+};
+
+const reject = () => {
```
polosanya, line 15: Same with reject

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+    message.textContent = 'Promise was rejected!';
+    document.body.append(message);
+  }, 3000);
+};
+
+const promise1 = new Promise(resolve);
+const promise2 = new Promise(reject);
+
+promise1.then();
+promise2.catch();
```
polosanya, line 29: These methods shouldn't be empty. Pass a function that appends a div with the text 'Promise was resolved/reject!' to the body.

polosanya (changes requested):
Almost done!

witflash (approved):
Well done 👍🏻

PR #603, opened 2023-07-06
solution

HannieA:
[DEMO LINK] https://HannieA.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,40 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function createMessage(message, isError = false) {
+  const messageDiv = document.createElement('div');
+
+  messageDiv.classList.add('message');
+
+  if (isError) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+  messageDiv.textContent = message;
+  document.body.appendChild(messageDiv);
+}
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', resolve);
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
```
lerastarynets, line 23: I think you misunderstood the task - `promise2` should be just rejected in 3 seconds and don't require a click, but that's not a big deal)

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

polosanya (approved):
Well done!

PR #604, opened 2023-07-06
solution

Gosyav:
https://vo7kov.github.io/js_promise_basic_DOM/


loralevitska (changes requested):
Almost perfect) Let's improve: now I can only see one  promise at a time, make two of them visible the first when clicking on the logo and the next after 3 seconds

Anastasiia-Svintsova (approved):

PR #605, opened 2023-07-17
add task solution

Luk2asz:
- [DEMO LINK](https://Luk2asz.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const div = document.getElementById('div');
+const logo = document.getElementById('logo');
+
+function handler() {
+  const promise1 = (resolve, reject) => {
+    logo.addEventListener('click', () => {
+      resolve('<div class="message">Promise was resolved!</div>')
+    });
+  };
+
+  return new Promise(promise1);
+}
+
+handler().then(function(result) {
+  div.innerHTML = result;
+});
+
+function handler2() {
+  const promise2 = (resolve, reject) => {
+    setTimeout(() => {
+      reject('<div class="message error-message">Promise was rejected!<div>');
+    }, 2999);
+  };
+
+  return new Promise(promise2);
+}
+
+handler2().catch(function(result) {
+  div.innerHTML = result;
+});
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+      resolve('<div class="message">Promise was resolved!</div>')
+    });
+  };
+
+  return new Promise(promise1);
+}
+
+handler().then(function(result) {
+  div.innerHTML = result;
+});
```
DorotaLeniecDev, line 18: Use arrow function as a callback
```suggestion
handler().then(result => {
  div.innerHTML = result;
});
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+function handler2() {
```
DorotaLeniecDev, line 20: Same here:
```const promise2 = new Promise((resolve, reject) => {// your code here}) ```

**Review by DorotaLeniecDev — CHANGES_REQUESTED** [review]
Check my comments

File: src/scripts/main.js [hunk 3]
```diff
@@ -1,3 +1,32 @@
...
+
+const addMessage = (text, className, className2) => {
+  const div = document.createElement('div');
+
+  document.body.append(div);
+  div.dataset.qa = 'notification';
+  div.classList.add(className);
+  div.classList.add(className2);
+  div.textContent = text;
+};
```
DorotaLeniecDev, line 14: Lets call it more descriptively, also you can pass multiple class name in one string:
```suggestion
const printMessage = (text, isError = false) => {
  const div = document.createElement('div');

  div.dataset.qa = 'notification';
  div.classList.add("message");
  isError && div.classList.add("error-message");
  div.textContent = text;
  document.body.append(div);
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+const promise1 = new Promise((resolve) => {
+  logo.addEventListener('click', () => {
+    // resolve('<div class="message">Promise was resolved!</div>');
```
DorotaLeniecDev, line 18: Remove comment

DorotaLeniecDev (changes requested):
2 comments from my side :)

PR #606, opened 2023-07-18
add task solution

maksym-nemera:
[DEMO LINK](https://maksym-nemera.github.io/js_promise_basic_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+      resolve(resolveMessage);
+    });
+
+    setTimeout(() => {
+      reject(rejectMessage);
+    }, 3000);
+  };
+
+  return new Promise(resolver);
+};
```
33Orange, line 29: In this case, you will create the same promise 2 times.
But they should be different.

image

33Orange (changes requested):
Fix linter errors, please.
|image

I think it got here by accident
image

PR #607, opened 2023-07-21
js_promise_basic_DOM

KarolinaWiktoria:
https://KarolinaWiktoria.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+
+const promise1 = new Promise((resolve) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    const resolveBlock = `
+    <div class="message">
+      Promise was resolved!
+    </div>`;
+
+    resolve(body.insertAdjacentHTML('beforeend', resolveBlock));
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  const rejectBlock = document.createElement('div');
+
+  rejectBlock.className = `error-message`;
+  rejectBlock.innerText = `Promise was rejected!`;
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+const promise1 = new Promise((resolve) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', () => {
+    const resolveBlock = `
+    <div class="message">
+      Promise was resolved!
+    </div>`;
+
+    resolve(body.insertAdjacentHTML('beforeend', resolveBlock));
```
DorotaLeniecDev, line 14: Try not to invoke function in a function here. Good idea would be to for example pass a message in resolve() function like:
```js
```

And then you could create a showMessage function and invoke it in the success handler:
```js
promise1.then(res => {
  showMessage(res);
});
```

`showMessage` function could be also used to display error message based on the parameters

DorotaLeniecDev (changes requested):
Use success and error handlers (then and catch) to invoke shoeMessage function which will be responsible for printing message in the document

src/scripts/main.js
KarolinaWiktoria: Okay, For sure ;). Thank you for your comments!

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
```
Radoslaw-Czerniawski, line 3: Body element is actually available on the document as property, no need for using the querySelector:
```js
const body = document.body;
```

Radoslaw-Czerniawski (approved):
All good :D, left one comment regarding the body.

src/scripts/main.js
KarolinaWiktoria: All right! Thank you for comments :)

PR #612, opened 2023-08-08
add task solution

ollavka:
[DEMO LINK](https://ollavka.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const body = document.body;
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
-// write your code here
+const logo = document.querySelector('.logo');
+const body = document.body;
+
+const createPromise = (withReject) => {
+  return new Promise((resolve, reject) => {
+    if (withReject) {
+      setTimeout(reject, 3000);
+
+      return;
```
DanilWeda, line 11: return se
```suggestion
     return setTimeout(reject, 3000);
```

PR #617, opened 2023-09-13
Add task solution

spojrzenie:


src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
 'use strict';
 
-// write your code here
+function promise1() {
```
darokrk, line 3: you can simplify this function to be just:

```
  document.querySelector('.logo').addEventListener('click', resolve);
});
```

and move logic for creating HTML in to separate function which will called on resolve status

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+      successDiv.classList.add('message');
+      successDiv.textContent = 'Promise was resolved!';
+      document.body.appendChild(successDiv);
+
+      resolve();
+    });
+  });
+}
+
+function promise2() {
```
darokrk, line 19: this second promise can be just: 

```
  setTimeout(() => {
    reject(onError);
  }, 3000);
});
```

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+      errorDiv.classList.add('message');
+      errorDiv.classList.add('error-message');
+      errorDiv.textContent = 'Promise was rejected!';
+      document.body.appendChild(errorDiv);
+
+      reject(new Error('Promise was rejected!'));
+    }, 3000);
+  });
+}
+
```
darokrk, line 33: You should create two function which will be call on resolve and reject status with created HTML elements, they can be, e.g:

```
const onSuccess = () => {
  document.body.insertAdjacentHTML('beforeend',
    `<div class="message">Promise was resolved!</div>`
  );
};

const onError = () => {
  document.body.insertAdjacentHTML('beforeend',
    `<div class="message error-message">Promise was rejected!</div>`
  );
};
```

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+promise1()
```
darokrk, line 34: here also we can now simplify our promises calls just to be:

```
promise1
  .then(onSuccess)
  .catch(onError);

promise2
  .then(onSuccess)
  .catch(onError);
```

we pass here our new created functions to methods .then and .catch

darokrk (changes requested):
Not bad 👏🏽  ! code works fine but can be simplified much further 👨🏽‍💻

src/scripts/main.js
spojrzenie: Thank you very much.

darokrk (approved):
Looks awesome, good job 🔥 👏🏽 !!

PR #639, opened 2023-11-19
solution

Timi2424:
- [DEMO LINK](https://<Timi2424>.github.io/js_promise_basic_DOM/)

I need explanation why then method for promise2 should be empty, or is it becouse the way the tests were written


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  document.querySelector('.logo').addEventListener('click', () => {
+    resolve('Promise 1 resolved!');
```
Barosz30, line 5: You dont realy need it here. It's realy added to site bit later.

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(new Error('Promise 2 rejected!'));
```
Barosz30, line 11: Same as above. There's no reason to use that here, if you put that to "message error-message" later.

Barosz30 (approved):
Overall, it looks ok. Tough i cant Promise. 

Pun intended

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+  .catch(() => {
+    const errorMessageDiv = document.createElement('div');
+
+    errorMessageDiv.className = 'message error-message';
+    errorMessageDiv.textContent = 'Promise was rejected!';
+    document.body.appendChild(errorMessageDiv);
+  });
+
+promise2
+  .then(() => {
```
mbruhler, line 32: You don't need this here if it will be rejected, just a catch method

src/scripts/main.js
mbruhler, line 11: you should just `reject()` or pass this argument later when you call this promise

src/scripts/main.js
mbruhler, line 5: You should just `resolve` or pass this argument later when you call this promise

mbruhler (changes requested):
There are couple of fixes to go.

Also work on styling a little bit, logo should be logo, and error / success should be also visibly changed

PR #640, opened 2023-11-20
initial commit - cant see tests

Barosz30:
I cant see test for some reason, but on site it works just as intented.

https://barosz30.github.io/js_promise_basic_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  const succesDiv = document.createElement('div');
+
+  succesDiv.className = 'message error-message';
+  succesDiv.innerText = 'Promise was rejected!';
+  body.appendChild(succesDiv);
+};
+
+promise1
+  .then(addMessage)
+  .catch(addBadMessage);
```
Patrykmclaren14, line 35: You can modify the `addMessage(value)` function here to sent an argument. This will make the function more versatile and able to handle different values.

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  setTimeout(() => reject(new Error()), 3000);
+});
+
+const addMessage = () => {
+  const succesDiv = document.createElement('div');
+
+  succesDiv.className = 'message';
+  succesDiv.innerText = 'Promise was resolved!';
+  body.appendChild(succesDiv);
+};
```
Patrykmclaren14, line 23: Here, you can modify the function to accept an argument and then append it as text

patrykyo20 (approved):
Task completed successfully, and you also added styles, so great job! My comments are just my perspective on the code, so they don't have to be correct. I hope I helped

mbruhler (approved):
As @Patrykmclaren14 stated, you can add the argument to the function that is passed to the promise handlers, but it's optional. So then you would be able to resolve with argument and this argument would be passed down to the DOM though this function. This is just FYI

Good job, approved.

PR #652, opened 2023-11-30
Develop

Vladimir-Zadorozhnyi:
https://Vladimir-Zadorozhnyi.github.io/js_promise_basic_DOM/


volodymyr-soltys97 (changes requested):
Hi, your demo link doesn't work, your need to make a deploy
image

PR #680, opened 2024-01-22
add task solution

AgnieszkaWw:
[DEMO LINK](https://AgnieszkaWw.github.io/js_promises_practice_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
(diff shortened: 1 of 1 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to fix the text as on the task description, also need to remove the third promise on the demo link
image

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review, you need yo update the demo link
image

PR #683, opened 2024-01-26
Promise solution

Sirius9312:
LINK: https://sirius9312.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+  });
+
+promise2
+  .catch(() => {
+    const rejectMessage = document.createElement('div');
+
+    rejectMessage.className = 'message error-message';
+    rejectMessage.textContent = 'Promise was rejected!';
+
+    document.querySelector('body').append(rejectMessage);
```
IvanFesenko, line 34: You can create helper function that will add message - it will keep code cleaner
```suggestion
});

const showMessage = (textContent , className ) => {
  const message = document.createElement('div');

  message.className = className ;
  message.textContent = textContent ;

  document.querySelector('body').append(message);
};

promise1
  .then(() => {
    showMessage('Promise was resolved!', 'message')
  });

promise2
  .catch(() => {
    showMessage('Promise was rejected!', 'message error-message')
```

IvanFesenko (approved):
Well done. Check comment

PR #709, opened 2024-03-23
Add solution

olafchuszno:
- [DEMO LINK](https://olafchuszno.github.io/js_promise_basic_DOM/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Create 2 promises `promise1` and `promise2`.
 
 ## Instructions
 1. Replace `<your_account>` with your Github username in the link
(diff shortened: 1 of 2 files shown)


yevhenii-pyl (changes requested):
Demo is not working, looks like there is no `gh-pages` branch at your repo and it was never deployed. Could you double-check?

olafchuszno:
Deployed now

yevhenii-pyl (approved):
Now we are talking, demo works as it should to 👍 
I would probably place a logic you have with creating DOM elements into success and failure functions, but can't see it as a reason to block the PR further :)

PR #717, opened 2024-03-31
Complete

KhomichYehor:
- [DEMO LINK](https://KhomichYehor.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,49 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const messageElement = document.createElement('div');
+
+messageElement.innerHTML = `Promise was resolved!`;
+
+messageElement.classList.toggle('message');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+      reject(new Error());
+    }, 3000);
+  });
+};
+
+const promise1 = firstPromise();
+
+const promise2 = secondPromise();
+
+function success() {
```
volodymyr-soltys97, line 39: It is important to follow the same way of creating functions everywhere
```suggestion
const success = () => {
```

volodymyr-soltys97 (approved):
Good job 👍

src/scripts/main.js
KhomichYehor: @volodymyr-soltys97 Thank you for your comments, in the future I will stick to one way of creating functions. Have a nice evening!

PR #723, opened 2024-04-18
Solution

meljaszuk:


src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+promise1
+  .then((data) => {
+    const div = document.createElement('div');
+
+    div.className = 'message';
+    div.innerText = data;
+
+    document.body.append(div);
+  })
+  .catch();
```
yevhenii-pyl, line 24: Interesting thing here: from one side, you always want to catch possible errors at any async operations to avoid `Uncaught Exception` if something goes wrong, however, from the other side, you want to keep your code as clean as possible. 
In this case, it makes sense to avoid adding `catch` since there is no chance for your promise to fail.

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+  });
+});
+
+const promise2 = new Promise(function (resolve, reject) {
+  setTimeout(() => reject(new Error('Promise was rejected!')), 3000);
+});
+
+promise1
+  .then((data) => {
+    const div = document.createElement('div');
```
yevhenii-pyl, line 17: You have almost identical logic for creating success and error messages. Why don't we create a helper function like 'generateMessage'?

PR #727, opened 2024-05-03
add task solution

DamianAdamczewski:
[DEMO LINK](https://DamianAdamczewski.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,42 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const messageElement = document.createElement('div');
+const errorMessageElement = document.createElement('div');
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+const promise2 = secondPromise();
+
+const success = () => {
+  document.body.appendChild(messageElement);
+};
+const failure = () => {
+  document.body.appendChild(errorMessageElement);
+};
+
+promise1.then(success).catch(failure);
```
yevhenii-pyl, line 41: Basically, no need to handle the error here since there never be an error, so just to keep your code cleaner .then will be enough.

yevhenii-pyl (approved):
LGTM 👍

PR #745, opened 2024-06-07
solution

VitaliiNez:
https://vitaliinez.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,52 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const promiseHandler = () => {
+  const promise1 = new Promise((resolve) => {
+    resolve('Promise was resolved!');
+  });
+
+  promise1
+    .then((response) => {
+      const divBlock = document.createElement('div');
+
+      divBlock.classList.add('message');
+
+      divBlock.textContent = response;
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,52 @@
...
+const promiseHandler = () => {
+  const promise1 = new Promise((resolve) => {
+    resolve('Promise was resolved!');
+  });
+
+  promise1
+    .then((response) => {
+      const divBlock = document.createElement('div');
+
+      divBlock.classList.add('message');
```
volodymyr-soltys97, line 13: Move this logic for add class to the helper variable and use it everywhere, it will be better

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+};
+
+const successHandler = (response) => {
+  const divBlock = document.createElement('div');
+
+  divBlock.classList.add('message');
+
+  divBlock.textContent = response;
+  document.body.append(divBlock);
+};
```
etojeDenys, line 21: you can try to combine these functions into one function

etojeDenys (approved):

PR #747, opened 2024-06-10
Solution

mukutiuk:
https://mukutiuk.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
+/* eslint-disable prefer-promise-reject-errors */
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
+/* eslint-disable prefer-promise-reject-errors */
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const div = document.createElement('div');
+
+div.classList.add('message');
```
etojeDenys, line 7: move all logic with creating the message to some function and just call that function in your then and catch

etojeDenys (changes requested):

vadiimvooo (approved):
Well done

PR #788, opened 2024-07-04
solution

iamgranovskaya:
[Demo](https://iamgranovskaya.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,51 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function createDiv(...classList) {
+  const div = document.createElement('div');
+
+  div.classList.add(...classList);
+
+  return div;
+}
+
+logo.addEventListener('click', (e) => {
+  const promise1 = new Promise((resolve, reject) => {
+    resolve('Promise was resolved!');
+  });
+
+  promise1
+    .then((data) => {
+      const message = createDiv('message');
+
+      message.innerHTML = data;
+      document.body.append(message);
+    })
+    .catch((error) => {
+      const message = createDiv('message', 'error-message');
+
+      message.innerHTML = error;
+      document.body.append(message);
+    });
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+      document.body.append(message);
+    });
+
+  const promise2 = new Promise((resolve, reject) => {
+    setTimeout(() => {
+      reject(new Error('Promise was rejected!'));
+    }, 3000);
+  });
+
+  promise2
```
etojeDenys, line 38: in your code, promise2 will be called only after the click(we don't have any note about it in the task description)

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+
+function createDiv(...classList) {
+  const div = document.createElement('div');
+
+  div.classList.add(...classList);
+
+  return div;
+}
+
+logo.addEventListener('click', (e) => {
```
etojeDenys, line 13: so, better to move that listener inside the first promise

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+      reject(new Error('Promise was rejected!'));
+    }, 3000);
+  });
+
+  promise2
+    .then((data) => {
+      const message = createDiv('message');
+
+      message.innerHTML = data;
+      document.body.append(message);
```
etojeDenys, line 43: move this logic to some common helper function and reuse it everywhere

etojeDenys (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+  message.innerHTML = data;
+  document.body.append(message);
+}
+
+function errorHandler(error) {
+  const message = createDiv('message', 'error-message');
+
+  message.innerHTML = error.message;
+  document.body.append(message);
+}
```
etojeDenys, line 25: actually you can try to create only one common function

etojeDenys (approved):
good job

PR #798, opened 2024-07-10
solution

isaeev:
[DEMO LINK](https://isaeev.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const message = document.createElement('div');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const message = document.createElement('div');
+
+message.classList.add('message');
+
```
etojeDenys, line 8: you should create the new blocks for each promise, so better just move this logic to some function

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #806, opened 2024-07-18
Develop

asushko25:
https://asushko25.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -10,7 +10,7 @@ const promise1 = new Promise((resolve, reject) => {
 
 const promise2 = new Promise((resolve, reject) => {
   setTimeout(() => {
-    reject('Error');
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+});
+
+promise1
+  .then((result) => {
+    const body = document.body;
+    const div = document.createElement('div');
+
+    div.classList.add('message');
+    div.textContent = 'Promise was resolved!';
+    body.appendChild(div);
```
etojeDenys, line 24: better to create some common function and move that logic there

and just call that function everywhere

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
GJ

PR #807, opened 2024-07-19
task solution

valost:
https://valost.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,41 @@
 'use strict';
 
-// write your code here
+function notifyAboutSuccess(message) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+}
+
+function notifyAboutError(message) {
+  const div = document.createElement('div');
+
+  div.classList.add('message', 'error-message');
+  div.textContent = message;
+
+  document.body.appendChild(div);
+}
```
etojeDenys, line 19: we can combine those functions into one

etojeDenys (approved):
lgtm

PR #818, opened 2024-07-22
solution

MariiaSuper:
https://MariiaSuper.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,5 +1,21 @@
 'use strict';
 
+const succsesFunc = (message) => {
+  const successDiv = document.createElement('div');
+
+  successDiv.className = 'message';
+  successDiv.innerText = message;
+  document.body.appendChild(successDiv);
+};
+
+const errorFunc = (error) => {
+  const errorDiv = document.createElement('div');
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+    }, 3000);
+  });
+
+  promise1
+    .then((message) => {
+      const successDiv = document.createElement('div');
+
+      successDiv.className = 'message';
+      successDiv.innerText = message;
+      document.body.appendChild(successDiv);
```
etojeDenys, line 23: better to create some common helper function and just call it in every then and catch

etojeDenys (changes requested):

anastasiiavorobiova (changes requested):
Make sure to deploy your solution

MariiaSuper:
> Make sure to deploy your solution



> Make sure to deploy your solution

I did it 20 hours ago (`npm run deploy`) but there are no actual deployment, I did ti couple minutes ago - no reaction, I rerun action - but it is queued

gh-pages did not not created as well

Somenting wrong with the deployment scripts (has a lot pf issues with the deployment on the another task)

image

PR #820, opened 2024-07-23
add task solution

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,48 @@
 'use strict';
 
-// write your code here
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+
+const promise1 = new Promise((resolve) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+  }, 3000);
+});
+
+promise1
+  .then(() => {
+    const message = document.createElement('div');
+
+    message.className = 'message';
+    message.textContent = 'Promise was resolved!';
+    body.appendChild(message);
```
volodymyr-soltys97, line 24: Create a helper function for this logic and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
```
volodymyr-soltys97, line 32: Use the helper `createMessage` here

volodymyr-soltys97 (changes requested):
Almost done!

volodymyr-soltys97 (approved):
Great 🔥

PR #822, opened 2024-07-23
Solution

Artem193:
https://Artem193.github.io/js_promise_basic_DOM/


lerastarynets (approved):
You're on the right track!

PR #829, opened 2024-07-30
add task solution

wasik07:
[DEMO LINK](https://wasik07.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const body = document.querySelector('body');
+
+const promise1 = new Promise(
+  (resolve) => (logo.onclick = () => resolve('Promise was resolved!')),
+);
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => reject(Error('Promise was rejected!')), 3000);
+});
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+  setTimeout(() => reject(Error('Promise was rejected!')), 3000);
+});
+
+promise1.then((message) => {
+  const div = document.createElement('div');
+
+  div.className = 'message';
+  div.textContent = `${message}`;
+
+  body.appendChild(div);
```
volodymyr-soltys97, line 20: Create a helper function for this logic and use it in two places

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+const body = document.querySelector('body');
+
+const promise1 = new Promise(
+  (resolve) => (logo.onclick = () => resolve('Promise was resolved!')),
```
etojeDenys, line 7: better to use addEventListener instead of html onclick

etojeDenys (approved):
good job

PR #832, opened 2024-07-31
solution

Radymyr:
[DEMO LINK](https://Radymyr.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,24 @@
 'use strict';
+/* eslint-disable */
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function promise1() {
+  logo.addEventListener('click', () =>
+    new Promise((resolve) => resolve('Promise was resolved!')).then(
+      (msg) => (document.body.innerHTML += `<div class="message">${msg}</div>`),
+    ),
+  );
+}
+
+function promise2() {
+  new Promise((resolve, reject) =>
+    setTimeout(reject, 3000, 'Promise was rejected!'),
+  ).catch(
+    (msg) =>
+      (document.body.innerHTML += `<div class="message error-message">${msg}</div>`),
+  );
+}
+
+promise1();
+promise2();
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Create 2 promises `promise1` and `promise2`.
 
 ## Instructions
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_promise_basic_DOM/)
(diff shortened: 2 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
 'use strict';
+/* eslint-disable */
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+function promise1() {
+  logo.addEventListener('click', () =>
+    new Promise((resolve) => resolve('Promise was resolved!')).then(
+      (msg) => (document.body.innerHTML += `<div class="message">${msg}</div>`),
```
TarasHoliuk, line 9: instead of adding inner HTML as string it's better to create a new div element, set the correct content and class for it, and append in the body

Such code would be easier to read and change (mantain)

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+    ),
+  );
+}
+
+function promise2() {
+  new Promise((resolve, reject) =>
+    setTimeout(reject, 3000, 'Promise was rejected!'),
+  ).catch(
+    (msg) =>
+      (document.body.innerHTML += `<div class="message error-message">${msg}</div>`),
```
TarasHoliuk, line 19: You're doing almost the same here

Let's create a separate function to add a new massage. As arguments it should accept the message and type (`error` || `success`) - to add the correct class

TarasHoliuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+    .catch((message) => {
+      div.classList.add('message', 'error-message');
+      div.textContent = message;
+      document.body.append(div);
+    });
+};
+
+showMessage(['Promise was rejected!', 'Promise was resolved!'], 'success');
+
+showMessage(['Promise was rejected!', 'Promise was resolved!'], 'error');
```
etojeDenys, line 32: create two different promises and only one common function for printing messages

etojeDenys (changes requested):
feel free to ask for some help with it in the chat

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+    .then((message) => {
+      div.classList.add('message');
+      div.textContent = message;
+      document.body.append(div);
+    })
+    .catch((message) => {
+      div.classList.add('message', 'error-message');
+      div.textContent = message;
+      document.body.append(div);
+    });
```
anastasiia-tilikina, line 26: it is better to resolve each promise separately not in function
make function that will be responsible for printing message only

anastasiia-tilikina (approved):
let's make some improvement

PR #833, opened 2024-08-01
done

VadymBryvka:


etojeDenys (changes requested):
no, you need to create two promises with `new Promise(........)` and call resolve/reject there after the click or timeout

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+
+  successMessage.textContent = 'Promise was resolved!';
+  resolve(successMessage);
+});
+
+successMessagePromise.then((successMessage) => {
+  document.querySelector('.logo').addEventListener('click', () => {
+    document.body.appendChild(successMessage);
+  });
+});
```
etojeDenys, line 17:
```suggestion
const successMessagePromise = new Promise((resolve) => {
  document.querySelector('.logo').addEventListener('click', () => {
    resolve('Promise was resolved!');
  });
});

successMessagePromise.then((successMessage) => {
  printMessage(successMessage);
});
```

you can also create the common print function and move logic with printing the message there, but the main point here that all async logic should be inside promises

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+  errorMessage.setAttribute('class', 'error-message message ');
+  errorMessage.textContent = 'Promise was rejected!';
+  reject(errorMessage);
+});
+
+erorPromise.catch((errorMessage) => {
+  setTimeout(() => {
+    document.body.appendChild(errorMessage);
+  }, 3000);
+});
```
etojeDenys, line 31: same here

etojeDenys (changes requested):

PR #845, opened 2024-08-08
solution

RomanPerepelkin:
https://RomanPerepelkin.github.io/js_promise_basic_DOM

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,31 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
+const promiseOne = new Promise((resolve) => {
+  logo.addEventListener('click', () => {
+    resolve();
+  });
+});
+
+const promiseTwo = new Promise((resolve, reject) => {
+  setTimeout(() => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+    reject(Error);
+  }, 3000);
+});
+
+promiseOne.then(() => {
+  const div = document.createElement('div');
+
+  div.classList.add('message');
+  div.innerHTML = 'Promise was resolved!';
+  document.body.appendChild(div);
```
etojeDenys, line 22: you can create some function with this logic and call it here

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+});
+
+promiseTwo.catch(() => {
+  const div = document.createElement('div');
+
+  div.className = 'message error-message';
+  div.innerHTML = 'Promise was rejected!';
+  document.body.appendChild(div);
```
etojeDenys, line 30: and here too

etojeDenys (approved):
good job

PR #850, opened 2024-08-10
task was solved

Pokhvalenkos:
[DEMO LINK](https://pokhvalenkos.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
 // write your code here
+const logo = document.querySelector('.logo');
+const body = document.querySelector('body');
+
+function createNotification(type, textMessage) {
+  const notification = document.createElement('div');
+
+  notification.className = type;
+  notification.textContent = textMessage;
+  body.append(notification);
+}
+
+logo.addEventListener('click', () => {
+  const promise1 = new Promise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+  notification.className = type;
+  notification.textContent = textMessage;
+  body.append(notification);
+}
+
+logo.addEventListener('click', () => {
+  const promise1 = new Promise((resolve, reject) => {
+    resolve();
+  });
+
```
etojeDenys, line 19:
```suggestion
const promise1 = new Promise((resolve, reject) => {
   logo.addEventListener('click', () => {
    resolve();
  }
});
```
should be listener inside the promise

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+const promise2 = new Promise(function (resolve, reject) {
+  reject(new Error());
+});
+
+promise2.catch(() => {
+  setTimeout(() => {
+    createNotification('error', 'Promise was rejected!');
+  }, 3000);
+});
```
etojeDenys, line 37: same here, move setTimeout inside the promise instead of catch

etojeDenys (changes requested):

PR #856, opened 2024-08-14
solution

Cy90sure:
[DEMO LINK](https://Cy90sure.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,33 @@
 'use strict';
 
-// write your code here
+function successHandler(text) {
+  const message = document.createElement('div');
+
+  message.className = 'message';
+  message.textContent = text;
+  document.body.append(message);
+}
+
+function errorHandler() {
+  const message = document.createElement('div');
+
+  message.className = 'message error-message';
+  message.textContent = 'Promise was rejected!';
+  document.body.append(message);
+}
+
+const promise1 = new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+  document.querySelector('.logo').addEventListener('click', () => {
+    resolve('Promise was resolved!');
+  });
+});
+
+promise1.then(successHandler).catch(errorHandler);
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    reject(Error);
```
TarasHoliuk, line 29:
```suggestion
    reject(new Error('Promise was rejected!'));
```
if you pass an error in the reject function you should pass its instance (i.e. create new Error)
In this case, you can read error.message in errorHandler

TarasHoliuk (changes requested):
1 small comment

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
-// write your code here
+function successHandler(text) {
+  const message = document.createElement('div');
+
+  message.className = 'message';
+  message.textContent = text;
+  document.body.append(message);
+}
+
+function errorHandler() {
```
TarasHoliuk, line 11:
```suggestion
function errorHandler(error) {
```
and then `message.textContent = error.message`. I meant you can read info from error
Just FYI

TarasHoliuk (approved):
GJ 👍 Approved

PR #857, opened 2024-08-15
solution

Fetrw:
[DEMO LINK](https://Fetrw.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -2,7 +2,7 @@
 
 const promise1 = new Promise((resolve) => {
   document.querySelector('.logo').addEventListener('click', () => {
-    resolve();
+    resolve('Promise was resolved!');
   });
 });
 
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve) => {
+  document.querySelector('.logo').addEventListener('click', () => {
+    resolve();
```
TarasHoliuk, line 5:
```suggestion
    resolve('Promise was resolved!');
```
and get it as an argument in the `handleSuccess` function to use as text content

TarasHoliuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+    reject(new Error('Promise was rejected!'));
+  }, 3000);
+});
+
+function handleSuccess(message) {
+  const messageDiv = document.createElement('div');
+
+  messageDiv.className = 'message';
+  messageDiv.textContent = message;
+  document.body.appendChild(messageDiv);
```
volodymyr-soltys97, line 20: Create a helper function for this logic and use it in two places

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

vadiimvooo (approved):
Well done

PR #862, opened 2024-08-19
Solution

lLiashko:
[DEMO LINK](https://lLiashko.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -29,9 +29,9 @@ const promise2 = new Promise((resolve, reject) => {
 });
 
 promise2
-  .then(() => {
-    addClass('message', 'Promise was resolved!');
+  .then((message) => {
+    addClass('message', message);
   })
-  .catch(() => {
-    addClass('message error-message', 'Promise was rejected!');
+  .catch((message) => {
+    addClass('message error-message', message);
   });
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -18,18 +18,10 @@ const promise1 = new Promise((resolve, reject) => {
 
 promise1
   .then(() => {
-    const messageDiv = document.createElement('div');
-
-    messageDiv.className = 'message';
-    messageDiv.innerHTML = 'Promise was resolved!';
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+  });
+});
+
+promise1
+  .then(() => {
+    const messageDiv = document.createElement('div');
+
+    messageDiv.className = 'message';
+    messageDiv.innerHTML = 'Promise was resolved!';
+    document.body.appendChild(messageDiv);
```
etojeDenys, line 17: you can move this logic to some helper function and reuse it everywhere

etojeDenys (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+  .catch(() => {
+    addClass('message error-message', 'Promise was rejected!');
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => reject(Error, 'Promise was rejected!'), 3000);
+});
+
+promise2
+  .then(() => {
```
TarasHoliuk, line 32:
```suggestion
  .then((messageText) => {
```
you can take it as an argument because you passed it to resolve: `resolve('Promise was resolved!');`

FYI

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+promise1
+  .then(() => {
+    addClass('message', 'Promise was resolved!');
+  })
+  .catch(() => {
+    addClass('message error-message', 'Promise was rejected!');
+  });
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => reject(Error, 'Promise was rejected!'), 3000);
```
TarasHoliuk, line 28:
```suggestion
  setTimeout(() => reject(new Error('Promise was rejected!')), 3000);
```
and then in catch callback get error object as an argument and read this text like this error.message
FYI

TarasHoliuk (approved):
Approved

Added 2 comments just FYI

PR #867, opened 2024-08-20
Develop

mrsvolodya:
https://mrsvolodya.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
+const logo = document.querySelector('.logo');
 
-// write your code here
+function createNotification(message, error = false) {
+  const div = document.createElement('div');
+    div.classList.add('message');
+
+  if (error) {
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+  if (error) {
+    div.classList.add('message', 'error-message');
+    div.textContent = message;
+    document.body.append(div);
+  } else {
+    div.textContent = message;
+    document.body.append(div);
+  }
+}
```
etojeDenys, line 18:
```suggestion
  div.classList.add('message');

  if (error) {
    div.classList.add('error-message');
  }
  
  div.textContent = message;
  document.body.append(div);
}
```

etojeDenys (changes requested):
looks like we can simplify it

lerastarynets (approved):
You're on the right track!

PR #869, opened 2024-08-22
done

v-zagorovskii:
- [DEMO LINK](https://v-zagorovskii.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,49 @@
 'use strict';
 
-// write your code here
+function createDiv() {
+  const div = document.createElement('div');
+
+  return div;
+}
+
+function createPositiveMessage(element) {
+  element.classList.add('message');
+  element.textContent = 'Promise was resolved!';
+  document.body.appendChild(element);
+}
+
+function createNegativeMessage(element, message) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
 'use strict';
 
-// write your code here
+function createDiv() {
+  const div = document.createElement('div');
+
+  return div;
+}
```
etojeDenys, line 7: why do we need this function, you can just create it directly document.createElement('div')

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+  element.classList.add('message');
+  element.textContent = 'Promise was resolved!';
+  document.body.appendChild(element);
+}
+
+function createNegativeMessage(element, message) {
+  element.classList.add('message', 'error-message');
+  element.textContent = message;
+  document.body.appendChild(element);
+}
```
etojeDenys, line 19: try to combine it in one common function

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #870, opened 2024-08-23
Develop

VikaChereushenko:
[DEMO LINK](https://VikaChereushenko.github.io/js_promise_basic_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -3,6 +3,15 @@
 // write your code here
 const body = document.querySelector('body');
 const logo = document.querySelector('.logo');
+const createElement = function (className, text) {
+  const div = document.createElement('div');
+
+  div.className = className;
+
+  div.textContent = text;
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
 'use strict';
 
 // write your code here
```
volodymyr-soltys97, line 3:
```suggestion

```

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+const body = document.querySelector('body');
+const logo = document.querySelector('.logo');
+const createElement = function (className, text) {
```
volodymyr-soltys97, line 6:
```suggestion
const createElement = (className, text) => {
```

volodymyr-soltys97 (approved):
Good job! 👍

PR #892, opened 2024-09-08
add task solution

modeltoIT:
[DEMO LINK](https://modeltoIT.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,31 @@
 'use strict';
 
-// write your code here
+const logo = document.querySelector('.logo');
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+});
+
+promise1.then((result) => {
+  const div = document.createElement('div');
+
+  div.classList.add('message');
+  div.textContent = 'Promise was resolved!';
+  document.body.append(div);
+
+  return result;
```
etojeDenys, line 20: we can move this logic to some function and reuse it everywhere

etojeDenys (approved):

PR #899, opened 2024-09-11
solved

Artem-Brui:
- [DEMO LINK](https://Artem-Brui.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -21,9 +21,7 @@ const promise2 = () => {
   });
 };
 
-promise2()
-  .then((res) => {})
-  .catch((mess) => createNewDiv(['message', 'error-message'], mess));
+promise2().catch((mess) => createNewDiv(['message', 'error-message'], mess));
 
 document.querySelector('.logo').addEventListener('click', () => {
   promise1().then((mess) => {
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -11,7 +11,9 @@ function createNewDiv(classList, textContent) {
 
 const promise1 = () => {
   return new Promise((resolve, reject) => {
-    resolve('Promise was resolved!');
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+};
+
+const promise2 = () => {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => reject(new Error('Promise was rejected!')), 3000);
+  });
+};
+
+promise2()
+  .then((res) => {})
```
etojeDenys:
```suggestion

```
you can remove it

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+
+    const existingDiv = [...divList].filter((div) => {
+      return div.className === 'message';
+    });
+
+    if (!existingDiv.length) {
+      createNewDiv(['message'], mess);
+    }
+  });
+});
```
etojeDenys: move the click event in the promise, and call resolve in click event callback

etojeDenys (changes requested):

src/scripts/main.js
Artem-Brui: thanks, resolved :)

src/scripts/main.js
etojeDenys: still not fixed
```
const promise1 = () => {
  return new Promise((resolve, reject) => {
    document.querySelector('.logo').addEventListener('click', () => {
    }
  });
};
```

etojeDenys (changes requested):

src/scripts/main.js
Artem-Brui: thanks, now is solved.

etojeDenys (approved):

PR #903, opened 2024-09-13
add task solution

DariaSydorenko:
- [DEMO LINK](https://DariaSydorenko.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,5 +1,14 @@
 'use strict';
 
+function createMessage(text, errorMessage) {
+  const message = document.createElement('div');
+
+  message.classList.add('message');
+  message.classList.add(errorMessage);
+  message.innerText = text;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+    reject(new Error('Error'));
+  }, 3000);
+});
+
+const success = () => {
+  const successMessage = document.createElement('div');
+
+  successMessage.classList.add('message');
+  successMessage.innerText = 'Promise was resolved!';
+  document.body.appendChild(successMessage);
```
volodymyr-soltys97: Create a helper function for this logic and use it in two places

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

etojeDenys (approved):

PR #908, opened 2024-09-16
add task solution

StanislavKapytsia:
- [DEMO LINK](https://StanislavKapytsia.github.io/js_promise_basic_DOM/)

Create 2 promises `promise1` and `promise2`.
- `promise1` should be resolved on a `click` on `.logo`.
- `promise2` promise should be rejected in 3 seconds.
- add success and error handlers for both promises
- success handler should append a `<div class="message">` with the text `Promise was resolved!` to the body
- error handler should append a `<div class="message error-message">` with the text `Promise was rejected!` to the body

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -61,25 +61,3 @@ promise2.catch((error, addInfo) => {
   }, 3000);
 });
 
-// const message = document.createElement('div');
-
-// message.classList.add('message');
-
-// const badResult = function (content) {
-//   message.classList.add('error-message');
-
-//   message.textContent = content;
-//   body.append(message);
-
-//   return message;
-// };
-
-// const goodResult = function (content) {
-//   message.classList.toggle('error-message', false);
-//   message.textContent = content;
-//   body.append(message);
-
-//   return message;
-// };
-
-
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -57,4 +57,3 @@ const promise2 = new Promise((resolve, reject) => {
 });
 
 promise2.catch((dataError) => getResult(dataError));
-
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -2,25 +2,20 @@
 
 const logo = document.querySelector('.logo');
 const body = document.querySelector('body');
-const message = document.createElement('div');
 
-message.classList.add('message');
+const getResult = function (content) {
+  const message2 = document.createElement('div');
 
-const badResult = function (content) {
-  message.classList.add('error-message');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,56 @@
...
+const goodResult = function (content) {
+  message.classList.toggle('error-message', false);
+  message.textContent = content;
+  body.append(message);
+
+  return message;
+};
+
+const promise1 = new Promise((resolve, reject) => {
+  logo.onclick = (e) => {
```
etojeDenys: in js, we usually use the addEventListener instead of the onClick

src/scripts/main.js
```diff
@@ -1,3 +1,56 @@
...
+  return message;
+};
+
+const goodResult = function (content) {
+  message.classList.toggle('error-message', false);
+  message.textContent = content;
+  body.append(message);
+
+  return message;
+};
```
etojeDenys: you need to create the message in this function instead of changing already created element, also try to combine these functions into one

etojeDenys (changes requested):

src/scripts/main.js
volodymyr-soltys97: Still not fixed 
image

volodymyr-soltys97 (changes requested):
Hi, you need to fix all comments from the previous review

src/scripts/main.js
```diff
@@ -1,3 +1,61 @@
...
+      }
+
+      data.message = `Promise was resolved!`;
+      resolve(data);
+    } else {
+      data.message = `Promise was rejected!`;
+      data.addInfo = 'error-message';
+
+      reject(data);
+    }
```
etojeDenys: why do you need this logic, you can just create the object in the resolve
```
resolve({
 ...........................
})
```
or in reject

etojeDenys (changes requested):

src/scripts/main.js
StanislavKapytsia: Знімок екрана 2024-09-17 165700

StanislavKapytsia: Знімок екрана 2024-09-17 165738

StanislavKapytsia: there is something problem with All checks have failed two 
the same code one ok another not

StanislavKapytsia: Знімок екрана 2024-09-17 173508
Знімок екрана 2024-09-17 173550

etojeDenys (approved):

PR #918, opened 2024-09-23
task solution

olshum8:
[DEMO LINK](https://olshum8.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve) => {
+  const button = document.querySelector('.logo');
+
+  button.addEventListener('click', () => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+  document.body.appendChild(successMessageDiv);
+}
+
+function errorHandler() {
+  const errorMessageDiv = document.createElement('vid');
+
+  errorMessageDiv.classList.add('message', 'error-message');
+  errorMessageDiv.textContent = 'Promise was rejected!';
+  document.body.appendChild(errorMessageDiv);
+}
```
etojeDenys, line 29: better to create one common function instead of two almost the same

etojeDenys (approved):

PR #938, opened 2024-10-08
Solution

sborichevskyi:
[DEMO LINK](https://sborichevskyi.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,49 @@
 'use strict';
 
-// write your code here
+const promise1 = new Promise((resolve, reject) => {
+  const logo = document.querySelector('.logo');
+
+  logo.addEventListener('click', (ev) => {
+    resolve();
+  });
+});
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    const error = 'Time is over';
+
+    reject(error);
+  }, 3000);
+});
+
+promise1
+  .then(() => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+  }, 3000);
+});
+
+promise1
+  .then(() => {
+    const messageDiv = document.createElement('div');
+
+    messageDiv.className = 'message';
+    messageDiv.textContent = 'Promise was resolved!';
+    document.body.appendChild(messageDiv);
```
volodymyr-soltys97: Create a helper function for this logic and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+
+const promise2 = new Promise((resolve, reject) => {
+  setTimeout(() => {
+    const error = 'Time is over';
+
+    reject(error);
+  }, 3000);
+});
+
+function makeDiv(isThen) {
```
volodymyr-soltys97, line 19:
```suggestion
function handleMessage(isThen) {
```

volodymyr-soltys97: Also, will be better if you pass messages and classes as props, not a boolean type

volodymyr-soltys97 (approved):
GJ!

PR #941, opened 2024-10-09
Solution

overnall:
- [DEMO LINK](https://overnall.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -15,18 +15,18 @@ const promise2 = new Promise((resolve, reject) => {
 });
 
 function successHandler() {
-  const div = document.createElement('div');
-
-  div.className = 'message';
-  div.innerText = 'Promise was resolved!';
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+    reject(new Error('Error'));
+  }, 3000);
+});
+
+function successHandler() {
+  const div = document.createElement('div');
+
+  div.className = 'message';
+  div.innerText = 'Promise was resolved!';
+  document.body.append(div);
```
volodymyr-soltys97: Сreate a helper function for this logic and use it in two places

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Great!

PR #942, opened 2024-10-09
add task solution

Syogunok:
- [DEMO LINK](https://Syogunok.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
-'use strict';
+const but = document.querySelector('.logo');
 
-// write your code here
+const prom1 = new Promise((resolve, reject) => {
+  but.addEventListener('click', () => {
+    const newDiv = document.createElement('div');
+
+    newDiv.innerHTML = 'Promise was resolved!';
+    newDiv.classList.add('message');
+
+    const parentElement = document.querySelector('body');
+
+    parentElement.appendChild(newDiv);
+
+    resolve();
+  });
+});
+
+prom1.then(() => {
+  return new Promise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


SanyaBratashchuk (changes requested):
second promise should be rejected regardless of whether first promise was resolver or not

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+const prom1 = new Promise((resolve) => {
+  but.addEventListener('click', () => {
+    const newDiv = document.createElement('div');
+
+    newDiv.innerHTML = 'Promise was resolved!';
+    newDiv.classList.add('message');
+
+    const parentElement = document.querySelector('body');
+
+    parentElement.appendChild(newDiv);
```
etojeDenys: you can simplify this logic

move all these lines to some helper function and call this function in `prom1.then(........)`

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+  setTimeout(() => {
+    const newDiv2 = document.createElement('div');
+
+    newDiv2.innerHTML = 'Promise was rejected!';
+    newDiv2.classList.add('message', 'error-message');
+
+    const parentElement = document.querySelector('body');
+
+    parentElement.appendChild(newDiv2);
+  }, 3000);
```
etojeDenys: do the same here, notice that you can create one common function for two promises

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #972, opened 2024-11-04
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/js_promise_basic_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -12,21 +12,14 @@ const promise2 = new Promise((resolve, reject) => {
   }, 3000);
 });
 
-const successHandler = (message) => {
+const showMessage = (raport) => {
   const messageDiv = document.createElement('div');
+  const errorInstance = raport instanceof Error;
 
-  messageDiv.className = 'message';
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+  messageDiv.className = 'message';
+  messageDiv.textContent = message;
+  document.body.appendChild(messageDiv);
+};
+
+const errorHandler = (error) => {
+  const messageDiv = document.createElement('div');
+
+  messageDiv.className = 'message error-message';
+  messageDiv.textContent = error.message;
```
zibi-dev: These functions perform the same actions with different class names and text content. I think we could create a utility function to keep the code DRY.

Zibi95 (changes requested):
Hey. Good job! One smal thing about DRY (Don't repeat yourself) principle to refactor and we are good to go.

Zibi95 (approved):
Good job!

PR #1177, opened 2025-03-27
finished

rekverr:


volodymyr-soltys97 (changes requested):
Hi, there is no demo link. Also, you need to provide demo link to the PR description 
image

